PR #1, opened 2024-08-23
[SECRES-2371] Adding logs to datadog

sobregosodd:


scfw/firewall.py
```diff
@@ -1,12 +1,16 @@
...
+import logging
 
 from scfw.cli import parse_command_line
 from scfw.commands import get_package_manager_command
 from scfw.target import InstallTarget
 from scfw.verifier import InstallTargetVerifier
 from scfw.verifiers import get_install_target_verifiers
+from scfw.config import LOG_DD
+
+log = logging.getLogger()
```
ikretz, line 13: Confirming my understanding: is the idea that this log is the "local" log and the one created on line 87 the log for just the things that get sent to Datadog?

And this logger created here on line 13 is the same as the one created in `config.py` by virtue of it being a singleton, right?

scfw/firewall.py
```diff
@@ -51,9 +55,9 @@ def verify_install_targets(verifiers: list[InstallTargetVerifier], targets: list
 
     with mp.Pool() as pool:
         pool.starmap(
-            _perform_verify_task,
-            [(verifier, target, findings) for verifier, target in verify_tasks]
-        )
+        _perform_verify_task,
+        [(verifier, target, findings) for verifier, target in verify_tasks]
+    )
```
ikretz: I think these are supposed to be indented one more level, no?

requirements.txt
```diff
@@ -1 +1,3 @@
 requests
+python-dotenv
+datadog-api-client
```
ikretz, line 3: Could you please also add these to the `pyproject.toml`?

scfw/firewall.py
```diff
@@ -79,30 +83,35 @@ def run_firewall() -> int:
...
             return 0
 
         command = get_package_manager_command(args.command, executable=args.executable)
-        if (targets := command.would_install()):
+        if targets := command.would_install():
+
             verifiers = get_install_target_verifiers()
+
             if (findings := verify_install_targets(verifiers, targets)):
+                tags = map(lambda x: f"{x.package}@{x.version}", findings.keys())
```
ikretz: You could use the `InstallTarget.show()` method here to format the package name and version.

I think also the `.keys()` call is not necessary, as the default iterator behavior of `dict` is over the keys.

scfw/firewall.py
```diff
@@ -79,30 +83,35 @@ def run_firewall() -> int:
...
+                ddlog.info(f"Instalation was block while attempting to run {command}", extra={"tags":[tags]})
```
ikretz: I think it should be "Installation was blocked..."

scfw/firewall.py
```diff
@@ -79,30 +83,35 @@ def run_firewall() -> int:
...
-            run_command = True
+                return 0
 
-        if run_command:
-            command.run()
+            if args.dry_run:
+                print("Exiting without installing, no issues found for installation targets.")
+                return 0
 
+        tags = map(lambda x: f"{x.package}@{x.version}", targets)
```
ikretz: Same thing here re `InstallTarget.show()`.

If it's simpler from a logging perspective, we could standardize `show()` so that it always formats targets the same way.  Right now, it's written to take ecosystem-specific formatting conventions into consideration.  For `npm`, it's `name@version` and for PyPI, they seem to prefer `name-version`.

scfw/firewall.py
```diff
@@ -79,30 +83,35 @@ def run_firewall() -> int:
...
-                run_command = True
-        else:
-            run_command = True
+                return 0
 
-        if run_command:
-            command.run()
+            if args.dry_run:
+                print("Exiting without installing, no issues found for installation targets.")
+                return 0
```
ikretz, line 108: I like the way you restructured the code so the early exits are right next to the conditions that create them.

scfw/firewall.py
sobregosodd: correct

scfw/firewall.py
sobregosodd: weird, i did not intend to change this code
apparently indentation in this case is just visual,  i'll re-apply the format.

scfw/firewall.py
sobregosodd: awesome.
usually you overrite the `__str__` method for that purpose, so you can easily do `str(obj)`

scfw/firewall.py
sobregosodd: np, i missed the `show()` method initially.

requirements.txt
sobregosodd: fixed in fc95ff08e5c0ca4e042c381b30cb96710d32815a

scfw/firewall.py
sobregosodd: fixed in fc95ff08e5c0ca4e042c381b30cb96710d32815a

scfw/firewall.py
sobregosodd: fixed in fc95ff08e5c0ca4e042c381b30cb96710d32815a

scfw/firewall.py
sobregosodd: fixed in fc95ff08e5c0ca4e042c381b30cb96710d32815a

scfw/firewall.py
sobregosodd: fixed in fc95ff08e5c0ca4e042c381b30cb96710d32815a

scfw/firewall.py
ikretz: Ahh, didn't know the trick about `__str__`.  I'll make that change.

PR #8, opened 2024-08-30
Add API documentation, GitHub Actions for code quality and tests

ikretz:
This PR adds support for building API documentation via `pdoc` as well as two GitHub Actions for enforcing code quality (type checking and linting) and testing on pull request and pushing to `main`.

Matrix testing is used to ensure compatibility with various versions of `pip` and `npm`.  This testing revealed that the firewall's current approach to parsing the output of `npm` dry-run commands is too dependent on this output having a certain format that is in fact only used in recent (i.e., `>=10.x`) versions of `npm`, precluding us from supporting more versions.  This will need to be addressed in a separate PR.  For now, a "disclaimer" has been added to the README regarding `npm` (and `pip`) compatibility, informed by these tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,4 +36,4 @@ coverage-report:
 docs:
 	pdoc --docformat google ./scfw > /dev/null &
 
-.PHONY: checks coverage test
+.PHONY: checks coverage install install-dev test
--- a/README.md
+++ b/README.md
@@ -31,8 +31,6 @@ add react 18.3.1
 added 3 packages in 127ms
 ```
 
-We hope to add support for older `npm` versions in the near future.
-
 Be advised that the firewall may fail to block installations of vulnerable or malicious packages if used with incompatible versions of `pip` or `npm`.
 
 ### Installation
@@ -78,7 +76,7 @@ The supply-chain firewall can optionally send logs of blocked and successful ins
 
 To opt in, set the environment variable `DD_API_KEY` to your Datadog API key, either directly in your shell environment or in a `.env` file in the current working directory.
 
-## Testing and development
+## Development
 
 To set up for testing and development, create a fresh `virtualenv`, activate it and run `make install-dev`.  This will install `scfw` and the development dependencies.
 
@@ -99,7 +97,7 @@ Of course, one can always test inside a container or VM for an added layer of pr
 
 The supply-chain firewall code may be typechecked with `mypy` and linted with `flake8`.  Run `make typecheck` or `make lint`, respectively, in the environment where the development dependencies have been installed.
 
-Run `make checks` to run the full suite of code quality checks, including tests.  These are the same checks that run in the firewall repository's CI, albeit with the tests run against various `pip` and `npm` versions to ensure compatibility.
+Run `make checks` to run the full suite of code quality checks, including tests.  These are the same checks that run in the repository's CI, the only difference being that the CI jobs matrix test against a range of `pip` and `npm` versions.  There is also a pre-commit hook that runs the checks in case one wishes to run them on each commit.
 
 ### Documentation
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -10,7 +10,19 @@ repos:
     name: Lint the firewall code
     language: system
     entry: make lint
-  - id: test
-    name: Run the test suite
+  - id: test-cli
+    name: Test the firewall CLI
     language: system
-    entry: make test
+    entry: make test-cli
+  - id: test-pip
+    name: Test firewall pip integration
+    language: system
+    entry: make test-pip
+  - id: test-npm
+    name: Test firewall npm integration
+    language: system
+    entry: make test-npm
+  - id: test-verifiers
+    name: Test firewall installation target verifiers
+    language: system
+    entry: make test-verifiers


requirements-dev.txt
```diff
@@ -1,5 +1,6 @@
 coverage
 flake8
 mypy
+pdoc
```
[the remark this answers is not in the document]
ikretz: We are using `pdoc 14.6.1` and not using math mode.  The project is unaffected by this issue.

Makefile
```diff
@@ -1,17 +1,39 @@
...
+	COVERAGE_FILE=.coverage.verifiers coverage run -m pytest tests/verifiers
 
 coverage-report:
+	coverage combine .coverage.cli .coverage.pip .coverage.npm .coverage.verifiers
 	coverage report
+
+docs:
+	pdoc --docformat google ./scfw > /dev/null &
+
+.PHONY: checks coverage test
```
sobregosodd: this runs `test` 2 times, one in this line and one called by coverage.
or makefile takes care of that?

README.md
```diff
@@ -1,21 +1,55 @@
 # Supply-chain firewall
 
+![Test](https://github.com/DataDog/supply-chain-firewall/actions/workflows/test.yaml/badge.svg)
+![Code quality](https://github.com/DataDog/supply-chain-firewall/actions/workflows/code_quality.yaml/badge.svg)
+
 This supply-chain firewall is a command-line tool for preventing the installation of vulnerable or malicious PyPI and npm packages.  It is intended primarily for use by engineers to protect their development workstations from compromise in a supply-chain attack.
 
+![scfw demo usage](images/demo.png)
```
sobregosodd, line 8: It would be cool to have a gif movie :)
i never did that before but it seems this tool works https://formulae.brew.sh/formula/ttygif

README.md
```diff
@@ -1,21 +1,55 @@
...
+```
+$ npm install --dry-run react
+add js-tokens 4.0.0
+add loose-envify 1.4.0
+add react 18.3.1
+
+added 3 packages in 127ms
+```
+
+We hope to add support for older `npm` versions in the near future.
```
sobregosodd: I would remove this

README.md
```diff
@@ -1,21 +1,55 @@
...
+
+Be advised that the firewall may fail to block installations of vulnerable or malicious packages if used with incompatible versions of `pip` or `npm`.
+
+### Installation
+
+Clone the repository, `cd` into the downloaded directory and run `make install`.  This will install the `scfw` command-line program into your global Python environment.  This can also be done inside a `virtualenv`, if desired.
```
sobregosodd, line 38: weird, i was expecting a `pip install .egg`

README.md
```diff
@@ -36,29 +70,21 @@ alias pip="scfw pip"
...
 
-The installation request was blocked.  No changes have been made.
-```
+![scfw datadog log](images/datadog_log.png)
 
-## Testing
+To opt in, set the environment variable `DD_API_KEY` to your Datadog API key, either directly in your shell environment or in a `.env` file in the current working directory.
 
-To run the test suite, first install `scfw` and the development dependencies.  It is recommended to create a fresh virtual environment for testing:
+## Testing and development
```
sobregosodd: I would just say development since Testing title is repeated below (but is referred to the test suite)

README.md
```diff
@@ -69,11 +95,15 @@ Exiting without installing, no issues found for installation targets.
 
 Of course, one can always test inside a container or VM for an added layer of protection, if desired.
 
-## Code quality
+### Code quality
```
sobregosodd, line 96: Do we need to mention the pre-commit tasks?

scfw/cli.py
```diff
@@ -1,16 +1,20 @@
...
 from argparse import ArgumentParser, Namespace
 import logging
 import sys
 
 from scfw.ecosystem import ECOSYSTEM
 
-LOG_LEVELS = list(map(logging.getLevelName, [logging.DEBUG, logging.INFO, logging.WARNING, logging.ERROR]))
-DEFAULT_LOG_LEVEL = logging.getLevelName(logging.WARNING)
+_LOG_LEVELS = list(map(logging.getLevelName, [logging.DEBUG, logging.INFO, logging.WARNING, logging.ERROR]))
+_DEFAULT_LOG_LEVEL = logging.getLevelName(logging.WARNING)
```
sobregosodd, line 12: you are over using  of underscore variables here.
i haven't seen that as a common practice, uppercased variables are expected to be treated as constants

scfw/command.py
```diff
@@ -14,20 +18,17 @@ def __init__(self, command: list[str], executable: Optional[str] = None):
         Initialize a new package manager command.
 
         Args:
-            self: The `PackageManagerCommand` to be initialized
             command: The package manager command line as provided to the supply-chain firewall.
-            executable: An optional path to the executable to use to run the package manager
-            commands. Must be determined from the environment if not provided.
+            executable:
```
sobregosodd, line 22: This executable is used only in pip case, right?
I mean, you cannot provide an npm binary I think. 
Is that confusing?

scfw/dd_logger.py
```diff
@@ -48,17 +64,17 @@ def emit(self, record):
...
+        os.environ["DD_SERVICE"] = _DD_SERVICE = _DD_APP_NAME
+    if not _DD_ENV:
+        os.environ["DD_ENV"] = _DD_ENV = "dev"
+    if not _DD_VERSION:
+        os.environ["DD_VERSION"] = _DD_VERSION = "0.1.0"
 
     ddlog = logging.getLogger(DD_LOG_NAME)
     ddlog.setLevel(logging.INFO)
-    ddlog_handler = DDLogHandler()
+    ddlog_handler = _DDLogHandler()
```
sobregosodd, line 77: Classes with underscore?

Makefile
ikretz: My understanding is that `.PHONY` modifies commands so that they are always rerun regardless of the file artifacts that may already exist.  But it doesn't run anything.

We should have `install` and `install-dev` under here, too, tbh.

README.md
ikretz: Yeah, I thought about that.  They look cool.  We can do it later.

README.md
ikretz: Why, is it standard?  I thought `pip install .` was simple and clear but I could change it.

README.md
ikretz: Yes, although with the CI, they're sort of optional now in case you want to run the checks every time you commit.  I'll update it.

scfw/cli.py
ikretz: The underscore prefix is a common Python convention for indicating which objects are part of the public API and which aren't.  These values are supposed to be private to the CLI module and could change at any time.

You'll see that, without the underscore, these constants appear in the API documentation.

scfw/command.py
ikretz: Sure, you could, it's just the path to your `npm`.  In fact, I think that's kind of even more common the `npm` side, as it's possible to have an installation of `npm` localized to a given project directory, no?

ikretz: That said, I expect it will be used almost exclusively on the Python side in practice.

scfw/dd_logger.py
ikretz: See previous comment about underscores.  I think this class should be considered internal.  In order to log something to Datadog, a client just needs the string `DD_LOG_NAME` to create a `getLogger` handle that uses this class internally.

README.md
ikretz: Agreed.

README.md
ikretz: Agreed.

scfw/dd_logger.py
ikretz: I should add that I might be overapplying this convention beyond what's standard in Python projects.  Coming from Rust and OCaml, there's a very rigid enforcement of public/private or external/internal objects, and I might be trying too hard to reproduce that here.

PR #11, opened 2024-09-03
[SECRES-2381] Support older versions of npm

### LINKED ISSUE #2: [SECRES-2381] Non-standard npm dry-run output formatting leading to false negatives
The supply-chain firewall currently depends on the output of `npm install --dry-run` having the following format: 

```
$ npm install --dry-run react
add js-tokens 4.0.0
add loose-envify 1.4.0
add react 18.3.1

added 3 packages in 406ms
```

This does not appear to be the case at least in version `9.5.0`.

We need to audit `npm` versions for the `--dry-run` output format, map out those versions that we currently support and those we do not, and make a decision about supporting the versions in this second category.

ikretz:
This PR reimplements the `NpmCommand.would_install()` method to make use of the `npm --loglevel silly` option. This was done to permit the supply-chain firewall to be compatible with older versions of `npm` whose `--dry-run` options are either completely silent or have different output formats than v10.x.  With this change, the firewall can support `npm` versions back to v7.0.

Closes #2

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,20 +18,9 @@ Current data sources used are:
 
 ### Compatibility
 
-The supply-chain firewall is compatible with `pip >= 22.2` and generally compatible with recent versions of `npm` (`>= 10.x`).
+The supply-chain firewall is compatible with `pip >= 22.2` and `npm >= 7.0`.  Before installing the supply-chain firewall, please upgrade to or verify that you are running a compatible version of `pip` or `npm`.
 
-In order to verify whether your `npm` is compatible, run an `npm install --dry-run` command for any package you do not already have installed and verify that the output resembles the following:
-
-```
-$ npm install --dry-run react
-add js-tokens 4.0.0
-add loose-envify 1.4.0
-add react 18.3.1
-
-added 3 packages in 127ms
-```
-
-Be advised that the firewall may fail to block installations of vulnerable or malicious packages if used with incompatible versions of `pip` or `npm`.
+Be advised that the firewall may fail to block installations of vulnerable or malicious packages if used with an incompatible version of `pip` or `npm`.
 
 ### Installation
 


scfw/commands/npm_command.py
```diff
@@ -23,12 +30,17 @@ def __init__(self, command: list[str], executable: Optional[str] = None):
             executable:
                 Optional path to the executable to run the command.  Determined by the
                 environment if not given.
+
+        Raises:
+            ValueError: An invalid `npm` command was given.
         """
-        assert command and command[0] == "npm", "Malformed npm command"
+        if not command or command[0] != "npm":
```
sobregosodd, line 37: great, i think we should avoid asserts

scfw/commands/npm_command.py
```diff
@@ -45,26 +57,45 @@ def would_install(self) -> list[InstallTarget]:
...
-            AssertionError: The `npm` dry-run output does not have the expected format.
+            ValueError: The `npm` dry-run output does not have the expected format.
         """
-        def is_add_line(line: str) -> bool:
-            return line.startswith("add") and not line.startswith("added")
+        def is_place_dep_line(line: str) -> bool:
+            return _NPM_LOG_PLACE_DEP in line
+
+        def line_to_dependency(line: str) -> str:
+            return line.split()[_NPM_LOG_DEP_TOKEN]
```
sobregosodd, line 66: this is always 4? no switch flag ever changes this line format?

scfw/commands/npm_command.py
```diff
@@ -45,26 +57,45 @@ def would_install(self) -> list[InstallTarget]:
...
+            dependencies = map(line_to_dependency, filter(is_place_dep_line, dry_run.stderr.split('\n')))
         except subprocess.CalledProcessError:
             # An error must have resulted from the given npm command
             # As nothing will be installed in this case, allow the command
             return []
+
+        try:
+            # List targets already installed in the npm environment
+            installed = subprocess.run([self._executable, "list"], check=True, text=True, capture_output=True).stdout
+        except subprocess.CalledProcessError:
```
sobregosodd, line 93: do we silently fail?
we should also have a flag to let the user decide if this the desired behaviour

scfw/firewall.py
```diff
@@ -42,13 +42,10 @@ def verify_install_targets(
...
     findings = {}
 
-    # Verify every target against every verifier
-    verify_tasks = itertools.product(verifiers, targets)
-
     with cf.ThreadPoolExecutor() as executor:
         task_results = {
             executor.submit(lambda v, t: v.verify(t), verifier, target): (verifier.name(), target)
-            for verifier, target in verify_tasks
+            for verifier, target in itertools.product(verifiers, targets)
```
sobregosodd, line 48: cool

scfw/commands/npm_command.py
ikretz: I realized too late that Python asserts are debug tools and the correctness of your code should not depend on them.  In fact, if you run Python in optimized mode, the asserts are removed.

scfw/commands/npm_command.py
ikretz: I looked at the `npm` source code for the latest v10 release and it's a fixed log line up to that point.  The first five tokens are guaranteed to be there.  I'll take a look for older npm versions before merging.

scfw/commands/npm_command.py
ikretz: Tbh I don't expect this situation to ever arise, since in order to get to this step where `npm list` fails, you already had to have successfully run the `--dry-run` command.  But yeah, maybe down the road we could provide a `--strict` option or something in case any error should cause the firewall to abort.

ikretz: But the current behavior on `npm list` error is correct from a safety standpoint, as it just means that all of the dependencies will be verified, even if some of them are already installed.

scfw/commands/npm_command.py
ikretz: Confirmed.

scfw/commands/npm_command.py
sobregosodd: can we at least add a log?

ikretz: Yes, good idea.  I'll roll it into the next PR.

PR #13, opened 2024-09-06
[SECRES-2397] Enforce pip and npm version compatibility

ikretz:
This PR changes the behavior of the firewall so that it refuses to run if an attempt is made to use it with an unsupported version of `pip` or `npm`.  This is in keeping with the firewall's primary goal of blocking 100% of vulnerable or malicious package installations within the purview of its datasets.

This is done by adding a custom exception, `UnsupportedVersionError`, which subclasses of `PackageManagerCommand` are intended to use to signal an incompatible version at initialization.  The firewall handles these errors gracefully, logging the error and exiting normally.  It will be very simple to change the code later on if we decide to let the user decide whether to run with an incompatible version in spite of this fact.

The PR also adds additional warning and info logs in the `PipCommand` and `NpmCommand` modules.

### DIFF
--- a/tests/commands/test_npm.py
+++ b/tests/commands/test_npm.py
@@ -13,7 +13,7 @@ npm_list = lambda : list_installed_packages(ECOSYSTEM.NPM)
 INIT_NPM_STATE = npm_list()
 TEST_TARGET = select_test_install_target(read_top_packages(TOP_NPM_PACKAGES), INIT_NPM_STATE)
 if not TEST_TARGET:
-    raise Exception("Unable to select target npm package for testing")
+    raise ValueError("Unable to select target npm package for testing")
 
 
 def test_npm_version_output():
--- a/tests/commands/test_pip.py
+++ b/tests/commands/test_pip.py
@@ -16,7 +16,7 @@ pip_list = lambda : list_installed_packages(ECOSYSTEM.PIP)
 INIT_PIP_STATE = pip_list()
 TEST_TARGET = select_test_install_target(read_top_packages(TOP_PIP_PACKAGES), INIT_PIP_STATE)
 if not TEST_TARGET:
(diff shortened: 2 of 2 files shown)


scfw/commands/pip_command.py
```diff
@@ -29,18 +38,34 @@ def __init__(self, command: list[str], executable: Optional[str] = None):
...
+            UnsupportedVersionError:
+                An unsupported version of `pip` was used to initialize a `PipCommand`.
         """
         def get_executable() -> str:
             if (venv := os.environ.get("VIRTUAL_ENV")):
                 return os.path.join(venv, "bin/python")
             else:
                 return sys.executable
 
+        def get_pip_version(executable: str) -> Version:
```
sobregosodd, line 50: I feel this can be a method `check_version()` of the abstract obj that returns bool instead of raising an error while getting the executable.
I think that such method can be reused in other parts of the code the future, but I don't have any strong opinion.

ikretz: Tbh what you're describing was my original vision for this change.  In the end, I decided to go with this error-based approach instead because I think it will overall lead to safer and more maintainable code.

In the `check_version()` style, to figure out whether the executable version is supported, you have to allow `PackageManagerCommand` instances for unsupported versions to exist.  But these latter instances have access to all of the other `PackageManagerCommand` methods that should only be executed when we have a supported version.  It means other parts of the code have to be much more careful about what kind of `PackageManagerCommand` instance it has, which probably means calling `check_version()` before proceeding with anything.

On the other hand, in the error-based approach, if you can instantiate the `PackageManagerCommand`, you know it's safe to use all of its methods.

The `check_version()` style has advantages, though, and if down the road we find ourselves needing to perform version checks in other parts of the code, we can change it.

PR #15, opened 2024-09-13
[SECRES-2352] Link to OSV.dev disclosures on block

ikretz:
This PR adds URLs to the OSV.dev disclosures for a blocked installation target so the user can get more information as to why the package is vulnerable or malicious.  As part of this change, all OSV.dev disclosures for a target (not just the first one with an ID) are displayed to the user.  It also makes the output on block more concise.

### DIFF
--- a/scfw/verifiers/osv_verifier.py
+++ b/scfw/verifiers/osv_verifier.py
@@ -57,9 +57,12 @@ class OsvVerifier(InstallTargetVerifier):
         request = requests.post(_OSV_DEV_QUERY_URL, json=query, timeout=10)
         request.raise_for_status()
 
-        if (vulns := request.json().get("vulns")):
-            osv_ids = filter(lambda id: id is not None, map(lambda vuln: vuln.get("id"), vulns))
-            message = f"An OSV.dev disclosure exists for package {target.show()}"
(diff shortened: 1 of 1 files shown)


scfw/verifiers/osv_verifier.py
```diff
@@ -56,12 +58,8 @@ def verify(self, target: InstallTarget) -> Optional[str]:
...
 
         if (vulns := request.json().get("vulns")):
-            osv_id = None
-            for vuln in vulns:
-                if (osv_id := vuln.get("id")):
-                    break
-            return (
-                f"An OSV.dev disclosure for package {target.package} exists (OSVID: {osv_id if osv_id else 'Unknown'})"
-            )
+            osv_ids = filter(lambda id: id is not None, map(lambda vuln: vuln.get("id"), vulns))
```
sobregosodd: you could avoid nesting her by inverting the condition:
```
if not (vulns := request.json().get("vulns")):
    return None
osv_ids = filter(lambda id: id is not None, map(lambda vuln: vuln.get("id"), vulns))
message = f"An OSV.dev disclosure exists for package {target.show()}"
return message + ''.join(map(lambda id: f"\n  * {_OSV_DEV_URL_PREFIX}/{id}", osv_ids))
```

ikretz: Nice, I'll change it.

PR #16, opened 2024-09-17
[SECRES-2431] Assign severities to findings, auto-block only on critical ones

### LINKED ISSUE #14: [SECRES-2431] Decide whether to restrict the firewall only to malicious packages
Currently, the supply-chain firewall's scope extends to vulnerable as well as malicious packages.  This is due to the fact that the firewall's OSV.dev package verifier considers that any disclosure relating to a package, not just `MAL` disclosures, constitutes an installation-blocking finding.

We have received feedback that this is not the expected behavior for the firewall: users expect only malicious (not vulnerable) packages to be blocked.  We need to decide whether to rescope the tool to match these expectations.

Generally, we have three options:

1. Do nothing: the firewall blocks on vulnerable as well as malicious packages
2. Reduce scope: the firewall only blocks on malicious packages and allows vulnerable ones
3. Compromise: Malicious packages are always blocked. Vulnerable packages are initially blocked but an option to continue the installation is provided

It should be noted that there is a non-insignificant number of OSV.dev disclosures for malicious packages that do not have a `MAL` identifier.

ikretz:
This PR changes the behavior of the firewall in two major ways:

1. It adds a simple hierarchy of severities, `WARNING` and `CRITICAL`, into the `InstallTargetVerifier` interface so that verifiers are required to rank their findings by severity.  Generally, `CRITICAL` is reserved for confirmed-malicious packages and (generally) `WARNING` is used for vulnerabilities or in cases where it is not clear that the finding means the target is malicious (e.g., as in the case of OSV.dev, in which some malicious package disclosures do not have a `MAL` ID).

2. Using these assigned severities, the firewall now auto-blocks only on `CRITICAL` findings and prompts the user for a confirmation of their intent to continue with the installation for `WARNING`-level findings.  An example of what this looks like can be seen in the README photo.

Various refactoring also went into supporting this change, including moving the orchestration layer for verifying targets into a separate module and creating a dedicated class to represent the results of this verification.

Closes #14

### DIFF
--- a/scfw/firewall.py
+++ b/scfw/firewall.py
@@ -42,7 +42,7 @@ def run_firewall() -> int:
 
         command = commands.get_package_manager_command(args.command, executable=args.executable)
         targets = command.would_install()
-        _log.info(f"Command would install: [{', '.join(t.show() for t in targets)}]")
+        _log.info(f"Command would install: [{', '.join(map(str, targets))}]")
 
         if targets:
             verifiers = verifs.get_install_target_verifiers()
@@ -55,18 +55,18 @@ def run_firewall() -> int:
             if (critical_report := reports.get(FindingSeverity.CRITICAL)):
                 dd_log.info(
                     f"Installation was blocked while attempting to run '{' '.join(args.command)}'",
-                    extra={"targets": map(lambda x: x.show(), critical_report.install_targets())}
+                    extra={"targets": map(str, critical_report.install_targets())}
                 )
-                print(critical_report.show())
+                print(critical_report)
                 print("\nThe installation request was blocked. No changes have been made.")
                 return 0
 
             if (warning_report := reports.get(FindingSeverity.WARNING)):
                 dd_log.info(
                     f"Seeking user confirmation while attempting to run '{' '.join(args.command)}'",
-                    extra={"targets": map(lambda x: x.show(), warning_report.install_targets())}
+                    extra={"targets": map(str, warning_report.install_targets())}
                 )
-                print(warning_report.show())
+                print(warning_report)
                 if _abort_on_warning():
                     print("The installation request was aborted. No changes have been made.")
                     return 0
@@ -76,7 +76,7 @@ def run_firewall() -> int:
             print("Dry-run: exiting without running command.")
         else:
             dd_log.info(
-                f"Running '{' '.join(args.command)}'", extra={"targets": map(lambda x: x.show(), targets)}
+                f"Running '{' '.join(args.command)}'", extra={"targets": map(str, targets)}
             )
             command.run()
         return 0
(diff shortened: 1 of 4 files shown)


scfw/firewall.py
```diff
@@ -117,18 +50,30 @@ def run_firewall() -> int:
                 f"Using installation target verifiers: [{', '.join(v.name() for v in verifiers)}]"
             )
 
-            if (findings := verify_install_targets(verifiers, targets)):
+            reports = verify.verify_install_targets(verifiers, targets)
+
+            if (critical_report := reports.get(FindingSeverity.CRITICAL)):
```
sobregosodd, line 55: feels a bit hardcoded

scfw/firewall.py
```diff
@@ -143,3 +88,18 @@ def run_firewall() -> int:
...
+
+def _abort_on_warning() -> bool:
+    """
+    Prompt the user for confirmation of whether or not to proceed with the
+    installation request in the case that there were `WARNING` findings.
+    """
+    try:
+        while (confirm := input("Proceed with installation? (y/N): ")) not in {'y', 'N', ''}:
+            pass
+        return confirm != 'y'
```
sobregosodd, line 101: what about doing lower()?

scfw/verify.py
```diff
@@ -0,0 +1,131 @@
...
+        Args:
+            target: The installation target the finding pertains to.
+            finding: The finding to be added to the report.
+        """
+        if target not in self._report:
+            self._report[target] = [finding]
+        else:
+            self._report[target].append(finding)
+
+    def show(self) -> str:
```
sobregosodd: you can use the `__str__()` trick here

ikretz: Here, I think I regret this decision to make this a class method.  It's really the `run_firewall()` function that wants to print things this way, not the `VerificationReport` class.  Before merging, I think I'm going to pull this method out and move it back into `firewall.py` as a function.  But yes, I do want to try the `__str__()` trick with `InstallTarget.show()` eventually.

scfw/firewall.py
ikretz: Good idea

scfw/firewall.py
ikretz: Yeah, I see what you mean.  Since we only have two severity levels, I think this is just the simple way to do it.

If we ever go back and add more levels, I think I would define a `FirewallAction` class and have a mapping from `FindingSeverity` to `FirewallAction`.

scfw/verify.py
ikretz: I decided to leave the method and use the `__str__()` trick.  I also did the same for `InstallTarget`.

PR #25, opened 2024-09-30
[SECRES-2430] Improve npm testing

ikretz:
This PR improves `npm` integration testing coverage by adding new GitHub Actions to exhaustively test the firewall against each `npm` minor version since v7.0.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -200,7 +200,7 @@ jobs:
           node-version: ${{ matrix.node-version }}
       - name: Test firewall npm v${{ matrix.npm-version }} integration
         run: |
-          if [ ${{ matrix.node-version }} = "15.0.0" ]; then
+          if [ "${{ matrix.node-version }}" = "15.0.0" ]; then
             # Known issue on npm v7.x: need to first install this manually
             npm install -g agentkeepalive
           fi


.github/workflows/test.yaml
```diff
@@ -72,12 +72,100 @@ jobs:
...
-        run: make test-npm
+          node-version: "15.0.0"
+      - name: Test firewall npm v7.x integration
+        run: |
+          # Need to install this separately, known issue with npm v7.x
+          npm install -g agentkeepalive
+          npm install -g npm@${{ matrix.npm-version }}
+          make test-npm
+
+  npm-integration-8x:
```
sobregosodd: Since the steps are the same you can reduce the code here having a [multidimensional matrix](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow#example-using-a-multi-dimension-matrix), you have 2 ways of doing that:

1. Extending the matrix for each npm version, [example](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow#expanding-or-adding-matrix-configurations)

2. more clean, create a dummy job that outputs an array of an array, and use the [output to define the matrices](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow#example-using-an-output-to-define-two-matrices). 
This might need some testing :-p

ikretz: I went with the first option for now because it's the simplest.  Not totally thrilled at how verbose it is, but if we want to, we can revisit it later.

PR #26, opened 2024-10-22
[SECRES-2420] Support custom verifiers and loggers

ikretz:
This PR adds support for automatically discovering installation target verifiers and loggers at runtime.  Users can now use custom verifiers and loggers with the firewall without having to modify its source code.

### DIFF
--- a/tests/commands/test_pip_command.py
+++ b/tests/commands/test_pip_command.py
@@ -47,16 +47,17 @@ def test_pip_command_would_install_exact():
         map(
             lambda p: InstallTarget(ECOSYSTEM.PIP, p[0], p[1]),
             [
-                ("certifi", "2024.8.30"),
-                ("charset-normalizer", "3.4.0"),
-                ("idna", "3.10"),
-                ("requests", "2.32.3"),
-                ("urllib3", "2.2.3")
+                ("botocore", "1.15.0"),
+                ("docutils", "0.15.2"),
+                ("jmespath", "0.10.0"),
+                ("python-dateutil", "2.9.0.post0"),
+                ("six", "1.16.0"),
+                ("urllib3", "1.25.11")
             ]
         )
     )
 
-    command_line = ["pip", "install", "--ignore-installed", "requests==2.32.3"]
+    command_line = ["pip", "install", "--ignore-installed", "botocore==1.15.0"]
     command = PipCommand(command_line, executable=sys.executable)
     targets = command.would_install()
     assert len(targets) == len(true_targets)


README.md
```diff
@@ -9,11 +9,13 @@ The supply-chain firewall is a command-line tool for preventing the installation
 
 The firewall collects all targets that would be installed by a given `pip` or `npm` command and checks them against reputable sources of data on open source malware and vulnerabilities.  The command is automatically blocked when any data source finds that any target is malicious.  In cases where a data source reports other findings for a target, the findings are presented to the user along with a prompt confirming intent to proceed with the installation.
 
-Current data sources used are:
+Default data sources include:
 
 - Datadog Security Research's public malicious packages [dataset](https://github.com/DataDog/malicious-software-packages-dataset)
 - [OSV.dev](https://osv.dev) disclosures
 
+Users may also implement verifiers for alternative data sources. Consult the API documentation for details.
```
sobregosodd: > Consult the API documentation for details.
This is a bit unclear, other products use a driven by example file or a folder structure and link it here.
For instance:
- `/examples/simple_logger.py` 
or 
- `custom_logger.py.example` at the root, the problem with this is that you have no syntax highlighting in github

scfw/loggers/__init__.py
```diff
@@ -1,16 +1,51 @@
...
+following name and signature:
+
+```
+def load_logger() -> FirewallLogger
+```
+
+This `load_logger` function should return an instance of the custom logger
+for the firewall's use. The module may then be placed in the same directory
+as this source file for runtime import. Make sure to reinstall the package
+after doing so.
```
sobregosodd, line 20: I would move this to an example file

scfw/loggers/__init__.py
```diff
@@ -1,16 +1,51 @@
...
+    Return the currently discoverable set of client loggers.
 
     Returns:
-        A `list` of the currently available `FirewallLogger`s.
+        A `list` of the discovered `FirewallLogger`s.
     """
-    return [DDLogger()]
+    loggers = []
+
+    for _, module, _ in pkgutil.iter_modules([os.path.dirname(__file__)]):
```
sobregosodd, line 42: mm... so, this is no right. I might be wrong but I see 2 problems here:

1. Aren't we importing the module every time  we call this function? That should no be the case.
2. Our app just uses `get_firewall_loggers` fx one time, so the problem is minimized but there is a reason why the python logger implements a singleton pattern, this is to prevent having multiple instances of the object whenever you want to log something, preventing memory drains.
Our `DDLogger` class is a non static class wrapping the singleton logger, thus, wasting the performance intended by the singleton.

My suggestion is implement a singleton here so if a DDLogger was previously instanciated we just return that instance, that will fix both problems

scfw/verifiers/__init__.py
```diff
@@ -1,22 +1,53 @@
...
+following name and signature:
+
+```
+def load_verifier() -> InstallTargetVerifier
+```
+
+This `load_verifier` function should return an instance of the custom verifier
+for the firewall's use. The module may then be placed in the same directory
+as this source file for runtime import. Make sure to reinstall the package
+after doing so.
```
sobregosodd, line 22: also move it to the examples folder / file

scfw/verifiers/__init__.py
```diff
@@ -1,22 +1,53 @@
...
-        available type.
+        A list of the discovered installation target verifiers.
     """
-    return [
-        DatadogMaliciousPackagesVerifier(),
-        OsvVerifier()
-    ]
+    verifiers = []
+
+    for _, module, _ in pkgutil.iter_modules([os.path.dirname(__file__)]):
```
sobregosodd, line 44: same here, do we want to reimport modules on every call?

tests/commands/test_pip_command.py
```diff
@@ -48,7 +48,7 @@ def test_pip_command_would_install_exact():
             lambda p: InstallTarget(ECOSYSTEM.PIP, p[0], p[1]),
             [
                 ("certifi", "2024.8.30"),
-                ("charset-normalizer", "3.3.2"),
+                ("charset-normalizer", "3.4.0"),
```
sobregosodd: did this changed? why?

README.md
ikretz: Good, will do

scfw/loggers/__init__.py
ikretz: How do you mean?

tests/commands/test_pip_command.py
ikretz: I think it's because 2.32.3 is the current version of `requests`, so they're still version-bumping their dependencies. This has happened before btw.

scfw/loggers/__init__.py
ikretz: Based on the [docs](https://docs.python.org/3/library/importlib.html#importlib.import_module), `importlib.import_module()` has the same semantics as the `import` statement, so a module will only be imported once unless you programatically `importlib.reload()` it.  So on the first point, I don't think we need to worry.

The singleton idea is a good one.  Let me look into it a bit and report back.

scfw/verifiers/__init__.py
ikretz: See previous comment.  For the verifiers, there is no need to implement singletons.

scfw/loggers/__init__.py
sobregosodd: i mean, this documentation on how write a custom logger can be moved to the example file i mentioned in the this comment https://github.com/DataDog/supply-chain-firewall/pull/26#discussion_r1814960466

scfw/loggers/__init__.py
ikretz: Added in [78b9f16](https://github.com/DataDog/supply-chain-firewall/pull/26/commits/78b9f1653c8c3a7832a6da0abf9299be566298c8)

scfw/loggers/__init__.py
ikretz: Oh, I see now, thanks.  I still want to have this be part of the `pdoc` API documentation, but I will document the example file, as well.

README.md
ikretz: Added in [1dfbd54](https://github.com/DataDog/supply-chain-firewall/pull/26/commits/1dfbd5447ca8d22372aab85826f3bda6da4f8c88)

scfw/loggers/__init__.py
ikretz: Added in [1dfbd54](https://github.com/DataDog/supply-chain-firewall/pull/26/commits/1dfbd5447ca8d22372aab85826f3bda6da4f8c88)

scfw/verifiers/__init__.py
ikretz: Added in [1dfbd54](https://github.com/DataDog/supply-chain-firewall/pull/26/commits/1dfbd5447ca8d22372aab85826f3bda6da4f8c88)

tests/commands/test_pip_command.py
sobregosodd: ahh, then I think this is generates a flaky test, because the test will fail every time they bump the version.
The goal of the test is to remain deterministic. 
Can we do something to lock what is resolved? Seems wrong doing so as well

scfw/loggers/dd_logger.py
```diff
@@ -118,3 +70,63 @@ def emit(self, record):
         with ApiClient(configuration) as api_client:
             api_instance = LogsApi(api_client)
             api_instance.submit_log(content_encoding=ContentEncoding.DEFLATE, body=body)
+
+
+# Configure a single logging handle for all `DDLogger` instances to share
+dotenv.load_dotenv()
```
sobregosodd, line 76: usually the `dotenv.load` imports several configs, looks weird importing the .env file in this class.
Maybe in the `__main__`? I don't think is a blocker, just a suggestion

tests/commands/test_pip_command.py
ikretz: Agreed, let's go back one minor version and try to freeze it there.

ikretz: So unfortunately the way they do their versioning on `requests` means we're going to keep running into this problem, even if we go to an earlier version.  We need to pick an entirely different package.

ikretz: I've looked among the top 20 packages and none of them lock their versions in such a way that this kind of thing will never happen, as far as I can tell.

I'll roll it back to the previous minor version of `requests` for now but we should probably just delete this test.  Or accept that we'll have to maintain it.

sobregosodd: ok, maybe we should just pick one very low on deps and with all deps pinned

sobregosodd: i feel the purpose of the test is not being met if we need to bump the packages every time.

ikretz: Let's delete it.

ikretz: Instead, I went to a very old version of `botocore` that seems very stable deps-wise.  If we keep having problems with this test, we can delete it.

scfw/loggers/dd_logger.py
ikretz: This is the only place `dotenv` is used, I think that call has been in here since we added this logger.

Since the loggers and verifiers are now auto-discovered, I think they should be self-contained: if this logger needs `dotenv`, it should be the one to call into it.

However, this could also be a guarantee we make to users, that we load from `dotenv` once before starting the run.  I think what we have is fine for now, but we can revisit it later if we find other loggers or verifiers reaching for `dotenv`.

PR #30, opened 2024-11-18
[SECRES-2564] Allow for controlling the level of Datadog logging

ikretz:
This PR exposes log levels for the Datadog logger to the user, allowing them to control through an environment variable what kinds of logs are sent to Datadog by the firewall.

By default, no logs are forwarded.  If they choose to opt-in, users may configure the logger so that only blocked installs, blocked + aborted installs, or all installs (blocked + aborted + allowed) are logged.

### DIFF
--- a/scfw/loggers/dd_logger.py
+++ b/scfw/loggers/dd_logger.py
@@ -91,11 +91,10 @@ class DDLogger(FirewallLogger):
         """
         self._logger = _ddlog
 
-        self._level = FirewallAction.BLOCK
         try:
             self._level = FirewallAction(os.getenv("SCFW_DD_LOG_LEVEL"))
         except ValueError:
-            pass
+            self._level = FirewallAction.BLOCK
 
     def log(
         self,


scfw/logger.py
```diff
@@ -16,8 +16,28 @@ class FirewallAction(Enum):
     package manager command.
     """
     Allow = 0
-    Block = 1
-    Abort = 2
+    Abort = 1
+    Block = 2
```
sobregosodd: Usually levels go in uppercase

README.md
```diff
@@ -76,7 +76,7 @@ The supply-chain firewall can optionally send logs of blocked and successful ins
 
 ![scfw datadog log](images/datadog_log.png)
 
-To opt in, set the environment variable `DD_API_KEY` to your Datadog API key, either directly in your shell environment or in a `.env` file in the current working directory.
+To opt in, set the environment variable `DD_API_KEY` to your Datadog API key, either directly in your shell environment or in a `.env` file in the current working directory.  A logging level must also be selected by setting the environment variable `SCFW_DD_LOG_LEVEL` to one of `Allow`, `Abort` or `Block`.  The `Block` level only logs blocked installations, `Abort` logs blocked and aborted installations, and `Allow` logs these as well as successful installations.
```
sobregosodd: what's the default?

scfw/loggers/dd_logger.py
```diff
@@ -90,7 +89,16 @@ def __init__(self):
         """
         Initialize a new `DDLogger`.
         """
-        self.logger = _ddlog
+        self._logger = _ddlog
+
+        self._level = None
```
sobregosodd: I'd say we be opinionated here, let's choose one by default if not defined. I'd say Block

scfw/loggers/dd_logger.py
```diff
@@ -90,7 +89,16 @@ def __init__(self):
...
+        match os.getenv("SCFW_DD_LOG_LEVEL"):
+            case "Allow":
```
sobregosodd: I suggest creating an enum and you can simply cast it or it will automatically raise if param not correct, like: 
```
DDLogLevel(Enum)
    BLOCK="BLOCK"
    ALLOW="ALLOW"
    
DDLogLevel(os.getenv("SCFW_DD_LOG_LEVEL", DDLogLevel.BLOCK)) 
```

README.md
ikretz: Default is no logging: users have to opt-in, no surprise logging charges.  But it sounds like being more opinionated is what you were thinking, so it could just as easily be "block."  There shouldn't be many logs of this type, anyway.

scfw/logger.py
ikretz: Technically, these are firewall actions, not levels.  I'm kind of overloading them as levels to do the Datadog logging.  I can capitalize them.

scfw/loggers/dd_logger.py
ikretz: Nice, I didn't know this was allowed.  Will change.

scfw/loggers/dd_logger.py
```diff
@@ -90,7 +89,13 @@ def __init__(self):
...
         Initialize a new `DDLogger`.
         """
-        self.logger = _ddlog
+        self._logger = _ddlog
+
+        self._level = FirewallAction.BLOCK
+        try:
+            self._level = FirewallAction(os.getenv("SCFW_DD_LOG_LEVEL"))
+        except ValueError:
+            pass
```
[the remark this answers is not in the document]
ikretz: If an exception occurs here, it is due to `SCFW_DD_LOG_LEVEL` being unset or incorrectly set.  In this case, the default behavior is to set the level to `BLOCK`, which has already been done, so the exception can be safely ignored.

scfw/logger.py
ikretz: Addressed in [3c0e720](https://github.com/DataDog/supply-chain-firewall/pull/30/commits/3c0e72021477446956a510c70e28f1d7ddc982cb)

README.md
ikretz: Updated in [3c0e720](https://github.com/DataDog/supply-chain-firewall/pull/30/commits/3c0e72021477446956a510c70e28f1d7ddc982cb)

scfw/loggers/dd_logger.py
ikretz: Addressed in [3c0e720](https://github.com/DataDog/supply-chain-firewall/pull/30/commits/3c0e72021477446956a510c70e28f1d7ddc982cb)

scfw/loggers/dd_logger.py
ikretz: Addressed in [3c0e720](https://github.com/DataDog/supply-chain-firewall/pull/30/commits/3c0e72021477446956a510c70e28f1d7ddc982cb)

PR #37, opened 2024-11-27
 [SECRES-2471] Fix logging to respect configured log level

### LINKED ISSUE #19: Datadog INFO logs always print regardless of log level
When Datadog logging is enabled, its `INFO`-level logs always print regardless of the log level selected by the user.

### LINKED ISSUE #35: Configure script fails to update correctly when all options are disabled.
The configure script takes no action when the user refuses all configuration options in the `scfw configure` script.  This means that if the user is trying to unset all configuration options, the script does not have the intended effect.  This needs to be taken into account.

### LINKED ISSUE #36: Datadog logging uses hardcoded default firewall version string
In `loggers/dd_logger.py`, the logger is configured to use a hard-coded version string of `0.1.0` when none is found environmentally.  This should always source the version value from the firewall's current version in `scfw.__version__`.

ikretz:
This PR adds a fix for a bug in the way the firewall's root logger was configured which resulted in the selected log level not being respected when Datadog logging was enabled, instead always printing the `INFO`-level logs from the Datadog logger.

It also fixes bugs in the `configure` script and in the sourcing of the firewall version in the Datadog logger.

Closes #19 
Closes #35 
Closes #36

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,12 @@ $ scfw configure
 
 ### Compatibility
 
-Supply-Chain Firewall is compatible with `pip >= 22.2` and `npm >= 7.0`.  In keeping with its goal of blocking 100% of known-malicious package installations, `scfw` will refuse to run with an incompatible version of `pip` or `npm`.  Please upgrade to or verify that you are running a compatible version of `pip` or `npm` before using this tool.
+|  Package manager  |  Compatible versions  |
+| :---------------: | :-------------------: |
+| npm               | >= 7.0                |
+| pip               | >= 22.2               |
+
+In keeping with its goal of blocking 100% of known-malicious package installations, `scfw` will refuse to run with an incompatible version of a supported package manager.  Please upgrade to or verify that you are running a compatible version before using this tool.
 
 Currently, Supply-Chain Firewall is only fully supported on macOS systems, though it should run as intended on most common Linux distributions.  It is currently not supported on Windows.
 


README.md
```diff
@@ -47,40 +51,40 @@ $ scfw configure
 
 ### Compatibility
 
-The supply-chain firewall is compatible with `pip >= 22.2` and `npm >= 7.0`.  In keeping with its goal of blocking 100% of known-malicious package installations, the firewall will refuse to run with an incompatible version of `pip` or `npm`.  Please upgrade to or verify that you are running a compatible version of `pip` or `npm` before using this tool.
+Supply-Chain Firewall is compatible with `pip >= 22.2` and `npm >= 7.0`.  In keeping with its goal of blocking 100% of known-malicious package installations, `scfw` will refuse to run with an incompatible version of `pip` or `npm`.  Please upgrade to or verify that you are running a compatible version of `pip` or `npm` before using this tool.
```
sobregosodd: Ideally this should be a table not text, like:

... some text:
Compatible with:

| Tool  |  Version |
|---|---|
|npm|>=7.0|

That way is more extensible and readable

scfw/cli.py
```diff
@@ -168,16 +168,15 @@ def _parse_command_line(argv: list[str]) -> tuple[Optional[Namespace], str]:
...
         args = parser.parse_args(argv[1:hinge])
 
-        # TODO(ikretz): Use `Subcommand` here instead of strings
         # Only allow a package manager `command` argument when
         # the user selected the `run` subcommand
-        match args.subcommand == "run", hinge == len(argv):
-            case True, False:
-                # `run` subcommand with `command` argument
+        match Subcommand(args.subcommand), argv[hinge:]:
+            case Subcommand.Run, []:
```
sobregosodd, line 174: Way clearer now! 👍

scfw/configure.py
```diff
@@ -141,7 +140,11 @@ def enclose(config: str) -> str:
     with open(config_file) as f:
         contents = f.read()
 
-    updated = re.sub(f"{_BLOCK_START}(.*?){_BLOCK_END}", enclose(config), contents, flags=re.DOTALL)
+    pattern = f"{_BLOCK_START}(.*?){_BLOCK_END}"
+    if not config:
+        pattern = f"\n{pattern}\n"
```
sobregosodd, line 145: why are you enclosing this with '\n' is that correct?

ikretz: Yeah, this turns out to be correct in testing (though there might be a better way of doing it).  You only add the extra `\n` in the case when you're completely deleting the firewall config section, which corresponds to/clears out the extra `\n` added in the very first config write.

README.md
ikretz: Changed in [55a069b](https://github.com/DataDog/supply-chain-firewall/pull/37/commits/55a069b02fa31a4d87a46d8c417cf4c34cfe6a8c)

PR #41, opened 2024-12-10
Allow configuration via command-line options

ikretz:
This PR adds command-line options to the `scfw configure` subcommand so that users may configure the command non-interactively via scripting.

```bash
usage: scfw configure [-h] [--alias-pip] [--alias-npm] [--dd-api-key KEY] [--dd-log-level LEVEL]

Configure the environment for using Supply-Chain Firewall.

options:
  -h, --help            show this help message and exit
  --alias-pip           Add shell aliases to always run pip commands through Supply-Chain Firewall
  --alias-npm           Add shell aliases to always run npm commands through Supply-Chain Firewall
  --dd-api-key KEY      API key to use when forwarding logs to Datadog
  --dd-log-level LEVEL  Desired logging level for Datadog log forwarding (options: ALLOW, ABORT, BLOCK)
```

If no command-line options are present, the interactive script runs as it did previously.  Otherwise, the command-line configuration is used and no output is printed.

The PR also performs various small refactoring.

### DIFF
--- a/scfw/commands/__init__.py
+++ b/scfw/commands/__init__.py
@@ -31,10 +31,13 @@ def get_package_manager_command(
     """
     if not command:
         raise ValueError("Missing package manager command")
-    match command[0]:
-        case ECOSYSTEM.PIP.value:
-            return ECOSYSTEM.PIP, PipCommand(command, executable)
-        case ECOSYSTEM.NPM.value:
-            return ECOSYSTEM.NPM, NpmCommand(command, executable)
-        case other:
-            raise ValueError(f"Unsupported package manager '{other}'")
+
+    try:
+        match (ecosystem := ECOSYSTEM(command[0])):
+            case ECOSYSTEM.PIP:
+                return ecosystem, PipCommand(command, executable)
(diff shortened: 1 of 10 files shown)


scfw/commands/__init__.py
```diff
@@ -31,10 +31,13 @@ def get_package_manager_command(
...
-    match command[0]:
-        case ECOSYSTEM.PIP.value:
-            return ECOSYSTEM.PIP, PipCommand(command, executable)
-        case ECOSYSTEM.NPM.value:
-            return ECOSYSTEM.NPM, NpmCommand(command, executable)
-        case other:
-            raise ValueError(f"Unsupported package manager '{other}'")
+
+    try:
+        match (ecosystem := ECOSYSTEM(command[0])):
```
sobregosodd, line 36: Usually enums are in camel case.

sobregosodd (approved):
LGTM

scfw/commands/__init__.py
ikretz: Yes, I wish I had done that to begin with.  Will roll this into the next major version.

PR #44, opened 2025-02-04
Add default logger for local Datadog Agent

ikretz:
This PR adds a new default logger to Supply-Chain Firewall that logs to the local Datadog Agent running on the user's system.  This logger can be enabled alongside the previous Datadog API logger, if desired, to send the same logs into different Datadog orgs.

It also updates the `scfw configure` command so that Agent logging can be configured interactively and non-interactively:
```bash
$ scfw configure --dd-agent-logging --dd-log-level ALLOW
```
Once this command is run, the user just needs to run `source ~/.bashrc` (or `.zshrc`) in their active shell to pick up the new configuration.  Otherwise, no additional action from the user is required to set up Agent logging.

### DIFF
--- a/scfw/configure.py
+++ b/scfw/configure.py
@@ -4,11 +4,14 @@ Implements the supply-chain firewall's `configure` subcommand.
 
 from argparse import Namespace
 import inquirer  # type: ignore
+import json
 import logging
 import os
 from pathlib import Path
 import re
+import subprocess
 import tempfile
+import time
 
 from scfw.logger import FirewallAction
 
@@ -23,6 +26,15 @@ DD_LOG_LEVEL_VAR = "SCFW_DD_LOG_LEVEL"
 """
 The environment variable under which the firewall looks for a Datadog log level setting.
 """
+
+DD_AGENT_LOG_VAR = "SCFW_DD_AGENT_LOGGING"
+"""
+The environment variable the presence of which the firewall uses to determine whether
+log forwarding to the local Datadog Agent has been enabled.
+"""
+
+DD_AGENT_PORT = 10518
+"""TCP port where the Datadog Agent receives logs from custom integrations"""
 
 _CONFIG_FILES = [".bashrc", ".zshrc"]
 
@@ -53,7 +65,9 @@ def run_configure(args: Namespace) -> int:
         An integer status code, 0 or 1.
     """
     try:
-        interactive = not any({args.alias_pip, args.alias_npm, args.dd_api_key, args.dd_log_level})
+        interactive = not any(
+            {args.alias_pip, args.alias_npm, args.dd_api_key, args.dd_log_level, args.enable_agent_logging}
+        )
 
         if interactive:
             print(_GREETING)
@@ -67,6 +81,9 @@ def run_configure(args: Namespace) -> int:
         for file in [Path.home() / file for file in _CONFIG_FILES]:
             if file.exists():
                 _update_config_file(file, _format_answers(answers))
+
+        if args.enable_agent_logging:
+            _configure_agent_logging()
 
         if interactive:
             print(_EPILOGUE)
@@ -142,6 +159,8 @@ def _format_answers(answers: dict) -> str:
         config += f'\nexport {DD_API_KEY_VAR}="{answers["dd_api_key"]}"'
     if answers["dd_log_level"]:
         config += f'\nexport {DD_LOG_LEVEL_VAR}="{answers["dd_log_level"]}"'
+    if answers["enable_agent_logging"]:
+        config += f'\nexport {DD_AGENT_LOG_VAR}="Enabled"'
 
     return config
 
@@ -174,3 +193,33 @@ def _update_config_file(config_file: Path, config: str) -> None:
     temp_handle.close()
 
     os.rename(temp_file, config_file)
+
+
+def _configure_agent_logging():
+    """
+    Configure a local Datadog Agent for accepting logs from the firewall.
+    """
+    config_yaml = (
+        "logs:\n"
+        "  - type: tcp\n"
+       f"    port: {DD_AGENT_PORT}\n"
+        '    service: "scfw"\n'
+        '    source: "scfw"\n'
+    )
+
+    agent_status = subprocess.run(["datadog-agent", "status", "--json"], check=True, text=True, capture_output=True)
+    agent_config_dir = json.loads(agent_status.stdout).get("config", {}).get("confd_path", "")
+
+    try:
+        scfw_config_dir = Path(agent_config_dir) / "scfw.d"
+        scfw_config_dir.mkdir()
+    except FileExistsError:
+        pass
+
+    with open(scfw_config_dir / "conf.yaml", 'w') as f:
+        f.write(config_yaml)
+
+    # Agent must be restarted for changes to take effect
+    subprocess.run(["launchctl", "stop", "com.datadoghq.agent"], check=True)
+    time.sleep(1)
+    subprocess.run(["launchctl", "start", "com.datadoghq.agent"], check=True)
--- a/scfw/configure.py
+++ b/scfw/configure.py
@@ -33,6 +33,8 @@ The environment variable under which the firewall looks for a port number on whi
 forward firewall logs to the local Datadog Agent.
 """
 
+_DD_AGENT_DEFAULT_LOG_PORT = "10365"
+
 _CONFIG_FILES = [".bashrc", ".zshrc"]
 
 _BLOCK_START = "# BEGIN SCFW MANAGED BLOCK"
@@ -115,19 +117,20 @@ def _get_questions() -> list[inquirer.questions.Question]:
         ),
         inquirer.Confirm(
             name="dd_agent_logging",
-            message="Would you like to enable sending firewall logs to your local Datadog Agent?",
+            message="If you have the Datadog Agent installed locally, would you like to forward firewall logs to it?",
             default=False
         ),
         inquirer.Text(
             name="dd_agent_port",
             message="Enter the local port where the Agent will receive logs",
+            default=_DD_AGENT_DEFAULT_LOG_PORT,
             ignore=lambda answers: not answers["dd_agent_logging"]
         ),
         inquirer.Confirm(
             name="dd_api_logging",
-            message="Would you like to enable sending firewall logs to Datadog?",
+            message="Would you like to enable sending firewall logs to Datadog using an API key?",
             default=False,
-            ignore=lambda _: has_dd_api_key
+            ignore=lambda answers: has_dd_api_key or answers["dd_agent_logging"]
         ),
         inquirer.Text(
             name="dd_api_key",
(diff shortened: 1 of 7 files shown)


scfw/configure.py
```diff
@@ -24,6 +27,23 @@
 The environment variable under which the firewall looks for a Datadog log level setting.
 """
 
+DD_AGENT_LOG_VAR = "SCFW_DD_AGENT_LOGGING"
```
sobregosodd: good call prepending the `SCFW_` 👍

scfw/configure.py
```diff
@@ -24,6 +27,23 @@
...
+"""
+The environment variable the presence of which the firewall uses to determine whether
+log forwarding to the local Datadog Agent has been enabled.
+"""
+
+DD_AGENT_PORT = 10518
```
sobregosodd: This one should also be configurable

scfw/configure.py
```diff
@@ -138,6 +168,8 @@ def _format_answers(answers: dict) -> str:
         config += '\nalias pip="scfw run pip"'
     if answers["alias_npm"]:
         config += '\nalias npm="scfw run npm"'
+    if answers["dd_agent_logging"]:
+        config += f'\nexport {DD_AGENT_LOG_VAR}="Enabled"'
```
sobregosodd: What about True or 1?, ie boolean casteable

scfw/configure.py
```diff
@@ -174,3 +206,37 @@ def enclose(config: str) -> str:
...
+    except subprocess.CalledProcessError:
+        raise RuntimeError("Unable to query Datadog Agent status. Ensure the Agent is running before retrying.")
+
+    scfw_config_dir = Path(agent_config_dir) / "scfw.d"
+    scfw_config_file = scfw_config_dir / "conf.yaml"
+    if scfw_config_file.is_file():
+        _log.info("Found existing Datadog Agent configuration file for Supply-Chain Firewall")
+        return
+
+    if not scfw_config_dir.is_dir():
```
sobregosodd: i would do a big try in case the user has no permissions to do this.

scfw/configure.py
```diff
@@ -174,3 +206,37 @@ def enclose(config: str) -> str:
...
     os.rename(temp_file, config_file)
+
+
+def _configure_agent_logging():
+    """
+    Configure a local Datadog Agent for accepting logs from the firewall.
+    """
+    try:
+        agent_status = subprocess.run(
+            ["datadog-agent", "status", "--json"], check=True, text=True, capture_output=True
```
sobregosodd: In the docker image the agent is called `agent`. 
```
# datadog-agent
bash: datadog-agent: command not found
```

scfw/configure.py
```diff
@@ -174,3 +206,37 @@ def enclose(config: str) -> str:
...
+        return
+
+    if not scfw_config_dir.is_dir():
+        scfw_config_dir.mkdir()
+    with open(scfw_config_file, 'w') as f:
+        f.write(_DD_AGENT_CONFIG_FILE)
+
+    # Agent must be restarted for changes to take effect
+    try:
+        subprocess.run(["launchctl", "stop", "com.datadoghq.agent"], check=True)
```
sobregosodd: This will only work in Mac, can we ask the agent to restart?
```suggestion
        subprocess.run(["agent", "stop"], check=True)
        subprocess.run(["agent", "run"], check=True)
```

scfw/loggers/dd_api_logger.py
```diff
@@ -0,0 +1,103 @@
...
+        super().__init__()
+
+    def emit(self, record):
+        """
+        Format and send a log to Datadog.
+
+        Args:
+            record: The log record to be forwarded.
+        """
+        if not (env := os.getenv("DD_ENV")):
```
sobregosodd: ain't better?: `os.getenv("DD_ENV","dev")`

scfw/loggers/dd_agent_logger.py
```diff
@@ -0,0 +1,100 @@
...
+_log = logging.getLogger(__name__)
+
+_DD_LOG_NAME = "dd_agent_log"
+
+
+class _DDLogFormatter(logging.Formatter):
+    """
+    A custom JSON formatter for firewall logs.
+    """
+    DD_ENV = "dev"
```
sobregosodd, line 24: hardcoded?

scfw/loggers/dd_api_logger.py
```diff
@@ -0,0 +1,103 @@
...
+        )
+
+        configuration = Configuration()
+        with ApiClient(configuration) as api_client:
+            api_instance = LogsApi(api_client)
+            api_instance.submit_log(content_encoding=ContentEncoding.DEFLATE, body=body)
+
+
+# Configure a single logging handle for all `DDAPILogger` instances to share
+dotenv.load_dotenv()
```
sobregosodd: Arent we already doing this in https://github.com/DataDog/supply-chain-firewall/pull/44/files#diff-800defce4b427b5b32759a584b90b13d24ab0d6f5593d4c1e1583a48de891249R73

sobregosodd (changes requested):

scfw/configure.py
ikretz: The [documentation](https://docs.datadoghq.com/agent/configuration/agent-commands/#agent-information) has it as `datadog-agent` for macOS.

scfw/loggers/dd_agent_logger.py
ikretz: This is the default value for when nothing is already set in the environment.

scfw/configure.py
ikretz: Based the [documentation](https://docs.datadoghq.com/agent/configuration/agent-commands/#start-the-agent), seems like it has to be through an external tool like `service` or `launchctl`.

scfw/configure.py
ikretz: I pulled this number right from the [docs](https://docs.datadoghq.com/agent/logs/?tab=tcpudp#custom-log-collection) thinking it was constant for custom integrations.  Something about the way it's written gives me this impression, but I see now that it must be configurable.  I will change this.

scfw/loggers/dd_api_logger.py
ikretz: I can do this, but I like leaving the default value in a constant

scfw/loggers/dd_api_logger.py
ikretz: We are, but these loggers are supposed to be independent of each other.  Plus, neither has priority over the other or is guaranteed to run, so they can't rely on each other to have run `load_dotenv()`.

They do now share a certain amount of code, let me see about moving it into there.

ikretz: Changed in [6f92c88](https://github.com/DataDog/supply-chain-firewall/pull/44/commits/6f92c8894d5be57ef70cc045d0e258d7e9a97b6d)

scfw/loggers/dd_api_logger.py
ikretz: Changed in [6f92c88](https://github.com/DataDog/supply-chain-firewall/pull/44/commits/6f92c8894d5be57ef70cc045d0e258d7e9a97b6d)

scfw/configure.py
ikretz: Added in [6f92c88](https://github.com/DataDog/supply-chain-firewall/pull/44/commits/6f92c8894d5be57ef70cc045d0e258d7e9a97b6d)

scfw/configure.py
ikretz: Changed in [6f92c88](https://github.com/DataDog/supply-chain-firewall/pull/44/commits/6f92c8894d5be57ef70cc045d0e258d7e9a97b6d)

scfw/configure.py
ikretz: I reversed this change when I remembered that the whole thing is wrapped in a big try-except (in `run_configure`).  The `try` blocks inside `_configure_agent_logging()` are just there to let me add a custom error message.

scfw/configure.py
ikretz: Added in [79cd5fd](https://github.com/DataDog/supply-chain-firewall/pull/44/commits/79cd5fdc2a9e27ed285edace421e9923a56ab608)

scfw/configure.py
sobregosodd: ok, so, are we in the clear that we are releasing a functionality in an OSS product only working for MacOS?

ikretz: We already state in the [README](https://github.com/DataDog/supply-chain-firewall?tab=readme-ov-file#compatibility) that the tool is only officially supported on macOS but that "it should run as intended on most common Linux distributions". This is still true except for this part of the configure script, so we could either take out the callout to Linux or just be more precise about what is compatible.

scfw/configure.py
```diff
@@ -100,7 +114,17 @@ def _get_questions() -> list[inquirer.questions.Question]:
             default=True
         ),
         inquirer.Confirm(
-            name="enable_dd_logs",
+            name="dd_agent_logging",
+            message="Would you like to enable sending firewall logs to your local Datadog Agent?",
+            default=False
+        ),
+        inquirer.Text(
+            name="dd_agent_port",
```
netcode, line 117: Can we set the default port to the question here?

scfw/configure.py
```diff
@@ -100,7 +114,17 @@ def _get_questions() -> list[inquirer.questions.Question]:
             default=True
         ),
         inquirer.Confirm(
-            name="enable_dd_logs",
+            name="dd_agent_logging",
+            message="Would you like to enable sending firewall logs to your local Datadog Agent?",
```
netcode:
```suggestion
            message="Do you have a local Datadog agent installed? Do you  like to enable sending firewall logs to it?",
```

scfw/configure.py
```diff
@@ -100,7 +114,17 @@ def _get_questions() -> list[inquirer.questions.Question]:
...
+            default=False
+        ),
+        inquirer.Text(
+            name="dd_agent_port",
+            message="Enter the local port where the Agent will receive logs",
+            ignore=lambda answers: not answers["dd_agent_logging"]
+        ),
+        inquirer.Confirm(
+            name="dd_api_logging",
```
netcode, line 122: I think we should ignore this question if they already answered yes to agent logging.

ikretz: Changed in [81b1349](https://github.com/DataDog/supply-chain-firewall/pull/44/commits/81b1349a59fdbf3cfdc4ac6a4ec4183eb25ed7da)

scfw/configure.py
ikretz: Changed in [81b1349](https://github.com/DataDog/supply-chain-firewall/pull/44/commits/81b1349a59fdbf3cfdc4ac6a4ec4183eb25ed7da)

scfw/configure.py
ikretz: Added in [81b1349](https://github.com/DataDog/supply-chain-firewall/pull/44/commits/81b1349a59fdbf3cfdc4ac6a4ec4183eb25ed7da)

sobregosodd (approved):

PR #46, opened 2025-02-07
Standardize Datadog log format

ikretz:
This PR makes it so that the Datadog API and Agent loggers use the same JSON-based log formatter.

It also updates the project's requirements files.


scfw/configure.py
```diff
@@ -16,6 +16,21 @@
...
+"""
+Source value for Datadog logging.
+"""
+
+DD_SERVICE = "scfw"
+"""
+Service value for Datadog logging.
+"""
+
+DD_ENV_DEFAULT = "dev"
```
sobregosodd: I find confusing this is called `_DEFAULT`

sobregosodd (approved):

PR #49, opened 2025-02-13
Add a configuration removal option

ikretz:
This PR adds a `--remove` option to the `scfw configure` subcommand that can be used to remove all SCFW-related configuration added to the environment via `scfw configure`, typically in preparation for uninstalling the tool.

This option unsets all shell aliases, removes SCFW-related environment variables, and deletes any Datadog Agent configuration added for log forwarding purposes.

### DIFF
--- a/scfw/cli.py
+++ b/scfw/cli.py
@@ -30,6 +30,13 @@ def _add_configure_cli(parser: ArgumentParser) -> None:
         parser: The `ArgumentParser` to which the `configure` command line will be added.
     """
     parser.add_argument(
+        "-r",
+        "--remove",
+        action="store_true",
+        help="Remove all configuration associated with Supply-Chain Firewall"
+    )
+
+    parser.add_argument(
         "--alias-pip",
         action="store_true",
         help="Add shell aliases to always run pip commands through Supply-Chain Firewall"
--- a/README.md
+++ b/README.md
@@ -61,6 +61,10 @@ In keeping with its goal of blocking 100% of known-malicious package installatio
 
 Currently, Supply-Chain Firewall is only fully supported on macOS systems, though it should run as intended on common Linux distributions.  It is currently not supported on Windows.
 
+### Uninstalling Supply-Chain Firewall
+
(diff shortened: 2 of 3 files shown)


scfw/cli.py
```diff
@@ -22,13 +22,20 @@
...
 
-def _add_configure_cli(parser: ArgumentParser) -> None:
+def _add_configure_cli(parser: ArgumentParser):
     """
     Defines the command-line interface for the firewall's `configure` subcommand.
 
     Args:
         parser: The `ArgumentParser` to which the `configure` command line will be added.
     """
+    parser.add_argument(
```
sobregosodd, line 32: ain't it better to add a group argument here to make remove and the adding arguments mutually exclusive?
if not possible i still think we need something to prevent both options so be used at the same time (to point it out to the user)

scfw/configure/__init__.py
```diff
@@ -0,0 +1,102 @@
...
+"""
+
+DD_AGENT_PORT_VAR = "SCFW_DD_AGENT_LOG_PORT"
+"""
+The environment variable under which the firewall looks for a port number on which to
+forward firewall logs to the local Datadog Agent.
+"""
+
+
+def run_configure(args: Namespace) -> int:
```
sobregosodd, line 47: why an int return? ain't it better a bool

sobregosodd (approved):

scfw/cli.py
ikretz: This should be possible.  I'll add it now if it's not too difficult, otherwise we can do it later.

scfw/configure/__init__.py
ikretz: This is the exit code that gets returned from `main::main()`

scfw/cli.py
ikretz: Added in [2742e7d](https://github.com/DataDog/supply-chain-firewall/pull/49/commits/2742e7dfc30092d1fce1f036b1d4769120359299)

It turns out that `argparse` doesn't have a way to define mutually exclusive argument groups.  But it was easy to add the condition to the internal parse function.

PR #53, opened 2025-03-05
Add executable to default log handler

ikretz:
This PR adds an `executable` field to the Datadog log handler that records the local filesystem path to the Python or npm executable that was used to execute the inspected package manager command.


scfw/command.py
```diff
@@ -30,6 +30,18 @@ def __init__(self, command: list[str], executable: Optional[str] = None):
...
+    def executable(self) -> str:
+        """
+        Query the local filesystem path to the executable that would be used to run the
+        package manager command.
+
+        Returns:
+            The local filesystem path of the executable that would be used to run the
+            package manager command in the current environment.
+        """
+        pass
```
sobregosodd, line 43: nit, aint it better a raise in the abstract?

scfw/commands/utils.py
```diff
@@ -0,0 +1,28 @@
...
+
+    Returns:
+        The local filesystem path to the given executable in the current environment
+        or `None` if the executable cannot be resolved.
+
+    Raises:
+        RuntimeError: Failed to resolve the given executable.
+    """
+    try:
+        path = subprocess.run(["which", executable], check=True, text=True, capture_output=True).stdout.strip()
```
sobregosodd: mind that which is not part of `coreutiils`.

scfw/command.py
sobregosodd: also, do we want to validate that the command is not injected, malformed or even exists?
I'm thinking that something like `scfw "npm something" install some-package` will break this

ikretz: Because of the way the inheritance is done with `metaclass=ABCMeta`, you can't instantiate a `PackageManagerCommand` if you don't implement all the methods so raising isn't necessary.

scfw/commands/utils.py
ikretz: Interesting.  This might be removed anyway as it's not working well with virtualenvs.  We'll see.

scfw/command.py
ikretz: Could you give me an example of the breakage you have in mind?

sobregosodd: try out feeding a non existing path for the executable

ikretz: Oh, I see.  We assume the user wants to run their command so has an incentive to get it right :) 
But since it's not a big deal, we could add validation for the executable path in `PipCommand` and `NpmCommand`.

scfw/commands/utils.py
ikretz: https://docs.python.org/3/library/shutil.html#shutil.which seems promising

ikretz: Removed in [b73af52](https://github.com/DataDog/supply-chain-firewall/pull/53/commits/b73af52e681058c708924720187ce54cc1d62593)

scfw/command.py
ikretz: Added `isfile` checks for the executables in [b73af52](https://github.com/DataDog/supply-chain-firewall/pull/53/commits/b73af52e681058c708924720187ce54cc1d62593)

PR #61, opened 2025-03-26
Rank OSV.dev findings by severity

ikretz:
This PR enhances the `OsvVerifier` so that its findings are ranked by severity:

```bash
$ scfw run pip install jinja2==3.1.4
Installation target Jinja2-3.1.4:
  - An OSV.dev disclosure exists for package Jinja2-3.1.4:
      * [High] https://osv.dev/vulnerability/GHSA-gmj6-6f8f-6699
  - An OSV.dev disclosure exists for package Jinja2-3.1.4:
      * [High] https://osv.dev/vulnerability/GHSA-q2x7-8rv6-6q7h
  - An OSV.dev disclosure exists for package Jinja2-3.1.4:
      * [Medium] https://osv.dev/vulnerability/GHSA-cpwx-vrp4-4pq7
```

Each OSV advisory optionally has a `severities` attribute containing a list of (usually CVSS) severity scores.  For each advisory returned for an install target, if this attribute is present, we compute the highest severity among these and sort the findings on this basis.  Findings for which no severity could be computed sort lower than findings with severities.

It should be noted that individual affected ranges are permitted to have their own severity scores by the OSV standard.  In such cases, the global `severities` attribute is not set.  For advisories like this, no severity score is currently assigned to the finding.

### DIFF
--- a/scfw/verifiers/osv_verifier/__init__.py
+++ b/scfw/verifiers/osv_verifier/__init__.py
@@ -3,6 +3,7 @@ Defines an installation target verifier that uses OSV.dev's database of vulnerab
 and malicious open source software packages.
 """
 
+import functools
 import logging
 
 import requests
@@ -103,9 +104,13 @@ class OsvVerifier(InstallTargetVerifier):
             mal_osvs = set(filter(lambda osv: osv.id.startswith("MAL"), osvs))
             non_mal_osvs = osvs - mal_osvs
 
+            osv_sort_key = functools.cmp_to_key(OsvAdvisory.compare_severities)
+            sorted_mal_osvs = sorted(mal_osvs, reverse=True, key=osv_sort_key)
+            sorted_non_mal_osvs = sorted(non_mal_osvs, reverse=True, key=osv_sort_key)
+
             return (
-                [(FindingSeverity.CRITICAL, finding(osv)) for osv in sorted(mal_osvs, reverse=True)]
-                + [(FindingSeverity.WARNING, finding(osv)) for osv in sorted(non_mal_osvs, reverse=True)]
+                [(FindingSeverity.CRITICAL, finding(osv)) for osv in sorted_mal_osvs]
+                + [(FindingSeverity.WARNING, finding(osv)) for osv in sorted_non_mal_osvs]
             )
 
         except requests.exceptions.RequestException as e:
--- a/scfw/verifiers/osv_verifier/__init__.py
+++ b/scfw/verifiers/osv_verifier/__init__.py
@@ -1,0 +1,129 @@
+"""
+Defines an installation target verifier that uses OSV.dev's database of vulnerable
+and malicious open source software packages.
+"""
+
+import logging
+
+import requests
+
+from scfw.ecosystem import ECOSYSTEM
+from scfw.target import InstallTarget
+from scfw.verifier import FindingSeverity, InstallTargetVerifier
+from scfw.verifiers.osv_verifier.osv_advisory import OsvAdvisory
+
+_log = logging.getLogger(__name__)
+
+_OSV_ECOSYSTEMS = {ECOSYSTEM.PIP: "PyPI", ECOSYSTEM.NPM: "npm"}
+
+_OSV_DEV_QUERY_URL = "https://api.osv.dev/v1/query"
+_OSV_DEV_VULN_URL_PREFIX = "https://osv.dev/vulnerability"
+_OSV_DEV_LIST_URL_PREFIX = "https://osv.dev/list"
+
+
+class OsvVerifier(InstallTargetVerifier):
+    """
+    An `InstallTargetVerifier` for the OSV.dev open source vulnerability and
+    malicious package database.
+    """
+    def name(self) -> str:
+        """
+        Return the `OsvVerifier` name string.
+
+        Returns:
+            The class' constant name string: `"OsvVerifier"`.
+        """
+        return "OsvVerifier"
+
+    def verify(self, target: InstallTarget) -> list[tuple[FindingSeverity, str]]:
+        """
+        Query an given installation target against the OSV.dev database.
+
+        Args:
+            target: The installation target to query.
+
+        Returns:
+            A list containing any findings for the given installation target, obtained
+            by querying for it against OSV.dev.
+
+            OSV.dev disclosures with `MAL` IDs are treated as `CRITICAL` findings and all
+            others are treated as `WARNING`.  *It is very important to note that most but
+            **not all** OSV.dev malicious package disclosures have `MAL` IDs.*
+
+        Raises:
+            requests.HTTPError:
+                An error occurred while querying an installation target against the OSV.dev API.
+        """
+        def mal_finding(osv: OsvAdvisory) -> str:
+            severity_tag = f"[{osv.severity}] " if osv.severity else ""
+            return (
+                f"An OSV.dev malicious package disclosure exists for package {target}:\n"
+                f"  * {severity_tag}{_OSV_DEV_VULN_URL_PREFIX}/{osv.id}"
+            )
+
+        def non_mal_finding(osv: OsvAdvisory) -> str:
+            severity_tag = f"[{osv.severity}] " if osv.severity else ""
+            return (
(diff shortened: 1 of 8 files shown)


scfw/verifiers/osv_verifier/osv_advisory.py
```diff
@@ -0,0 +1,217 @@
...
+
+        Raises:
+            TypeError: The other argument given was not a `Severity`.
+        """
+        if self.__class__ is not other.__class__:
+            raise TypeError(
+                f"'<' not supported between instances of '{self.__class__}' and '{other.__class__}'"
+            )
+
+        return (self.name, other.name) in {
```
tesnim5hamdouni: Mapping the severity levels to numerical values is more straight-forward 
Then in __str__ you introduce the string values

scfw/verifiers/osv_verifier/osv_advisory.py
```diff
@@ -0,0 +1,217 @@
...
+            or greater than the second one.
+
+        Raises:
+            TypeError: One of the given arguments is not an `OsvAdvisory`.
+        """
+        if not (isinstance(lhs, cls) and isinstance(rhs, cls)):
+            raise TypeError("Received incompatible argument types while comparing OSV severities")
+
+        # A match statement would be more natural here but mypy is not up to it
+        if lhs.severity == rhs.severity:
```
tesnim5hamdouni, line 161: if a numerical mapping is chosen, this part too can be simplified

tesnim5hamdouni (approved):
Change pending

scfw/verifiers/osv_verifier/osv_advisory.py
ikretz: Changed in [97266e2](https://github.com/DataDog/supply-chain-firewall/pull/61/commits/97266e2d5f6263fb140860eb76a709f39d8240ee)

tesnim5hamdouni (approved):

scfw/verifiers/osv_verifier/osv_advisory.py
ikretz: Here, I think we need this logic because `self.severity` can be `None`, so the first three cases are needed for eliminating that possibility.  Unless you see a simpler way?

scfw/verifiers/osv_verifier/__init__.py
```diff
@@ -53,16 +55,12 @@ def verify(self, target: InstallTarget) -> list[tuple[FindingSeverity, str]]:
             requests.HTTPError:
                 An error occurred while querying an installation target against the OSV.dev API.
         """
-        def mal_finding(id: str) -> str:
+        def finding(osv: OsvAdvisory) -> str:
```
sobregosodd, line 58: It feel like this function can be a method of `OsvAdvisory`, something like, `to_disclusure_str()`

scfw/verifiers/osv_verifier/osv_advisory.py
```diff
@@ -0,0 +1,212 @@
...
+    """
+    Represents the possible severities that an OSV advisory can have.
+    """
+    Non = 0
+    Low = 1
+    Medium = 2
+    High = 3
+    Critical = 4
+
+    def __lt__(self, other: "Severity") -> bool:
```
sobregosodd: TIL
You can also use Self

scfw/verifiers/osv_verifier/osv_advisory.py
```diff
@@ -0,0 +1,212 @@
...
+        return self.value < other.value
+
+    def __str__(self) -> str:
+        """
+        Format a `Severity` for printing.
+
+        Returns:
+            A `str` representing the given `Severity` suitable for printing.
+        """
+        return "None" if self.name == "Non" else self.name
```
sobregosodd, line 51: Aint better to use the value in the if? 

```suggestion
        return "None" if self.value == self.Non else self.name
```

Also, why not just change the enum: Non -> None

scfw/verifiers/osv_verifier/osv_advisory.py
```diff
@@ -0,0 +1,212 @@
...
+        Returns:
+            The `OsvSeverityType` referred to by the given string.
+
+        Raises:
+            ValueError: The given string does not refer to a valid `OsvSeverityType`.
+        """
+        mappings = {type.value: type for type in cls}
+        if (type := mappings.get(s)):
+            return type
+        raise ValueError(f"Invalid OSV severity type '{s}'")
```
sobregosodd: if enum values are str, you don't need to do this, you can just do multiple inheritance of `str` and `Enum` (in that order). Then you can `OsvSeverityType("CVSS_V4")`

sobregosodd (approved):
some nits

scfw/verifiers/osv_verifier/osv_advisory.py
ikretz: I was trying to use `Self` from `typing-extensions` but wasn't able to get `mypy` to accept it.  Apparently `typing` added `Self` in v3.11 but we're on v3.10 and up. 🤷  I'll try again to get it working.

scfw/verifiers/osv_verifier/osv_advisory.py
ikretz:
```python
>>> from scfw.verifiers.osv_verifier.osv_advisory import Severity
Traceback (most recent call last):
...
    None = 0
    ^^^^
SyntaxError: cannot assign to None
```

Tbh this shouldn't be an issue because this None is inside the Severity namespace. 🤷

scfw/verifiers/osv_verifier/osv_advisory.py
ikretz: TIL

scfw/verifiers/osv_verifier/osv_advisory.py
ikretz: I can use the value in the return statement.

sobregosodd: oh, right, is a reserved word

scfw/verifiers/osv_verifier/osv_advisory.py
sobregosodd: nah, your approach works as well... don't waste time

scfw/verifiers/osv_verifier/osv_advisory.py
ikretz: After thinking on this one a bit more, I think I will leave it the way it is using `self.name` as the criterion.  The variant names are defined in the CVSS standard and thus won't be changed whereas the numbers are arbitrary.  This implementation also communicates intent slightly better (i.e., I wanted it to be `self.name` but `None` is reserved, so I have to do this thing).

scfw/verifiers/osv_verifier/osv_advisory.py
ikretz: Updated in [68c6e83](https://github.com/DataDog/supply-chain-firewall/pull/61/commits/68c6e839e28257c5eeef344d2aeb8791d4faa65e)

scfw/verifiers/osv_verifier/osv_advisory.py
ikretz: Updated in [68c6e83](https://github.com/DataDog/supply-chain-firewall/pull/61/commits/68c6e839e28257c5eeef344d2aeb8791d4faa65e)

scfw/verifiers/osv_verifier/__init__.py
ikretz: On this one, I completely agree, but it would involve a lot of detailed extra work to parse the OSV JSON `affected` field correctly.  Unfortunately for us, OSV advisories are not in a 1-1 relationship with install targets or even packages, since a single advisory can cover multiple packages (even in different ecosystems).

The alternative is just to not mention the install target by name in the finding.  I.e., `"An OSV.dev disclosure exists for this installation target"` instead of `"An OSV.dev disclosure exists for package X"`.  But I feel like this makes the advisories less informative and less useful.

I say let's not change it for now.  One day, I will write a proper OSV parser and we can refactor.

PR #74, opened 2025-04-22
Add initial support for Poetry

ikretz:
This PR adds initial support for the PyPI ecosystem [Poetry](https://python-poetry.org/) package manager.

Supported subcommands:
* `poetry add`

Any other Poetry commands are automatically allowed to ensure compatibility.  Poetry >=1.3.0 is supported.

The `scfw configure` subcommand has also been updated to accommodate adding shell aliases for `poetry`.

Finally, minor refactoring was done in the `PackageManagerCommand` class, Poetry-specific tests were added, and dependencies were updated.


tesnim5hamdouni (approved):
Waiting for Sebastian's input

README.md
```diff
@@ -45,27 +45,30 @@ $ scfw --version
...
+### Compatibility and limitations
 
-|  Package manager  |  Compatible versions  |
-| :---------------: | :-------------------: |
-| npm               | >= 7.0                |
-| pip               | >= 22.2               |
+|  Package manager  |  Compatible versions  |     Supported subcommands     |
+| :---------------: | :-------------------: | :---------------------------: |
+| npm               | >= 7.0                | `install` (including aliases) |
+| pip               | >= 22.2               | all                           |
```
sobregosodd: this makes it look like the you can use all pip commands but instead it will only work with install in npm.
it reads odd, maybe we should rename the column to `Actions on subcomands` but again, we don't action on all pip commands, no?

README.md
```diff
@@ -45,27 +45,30 @@ $ scfw --version
...
+| poetry            | >= 1.3                | `add`                         |
 
 In keeping with its goal of blocking 100% of known-malicious package installations, `scfw` will refuse to run with an incompatible version of a supported package manager.  Please upgrade to or verify that you are running a compatible version before using this tool.
 
+Supply-Chain Firewall may only know how to instrument some of the "installish" subcommands for its supported package managers.  Supported subcommands are shown in the above table.  We hope to extend the tool's purview to all relevant subcommands for each package manager over time.  *Unsupported installish subcommands are always allowed*.
```
sobregosodd: > We hope to extend the tool's purview to all relevant subcommands for each package manager over time

IMO this is too verbose for a README, it can bloat the reading, maybe we should add a public roadmap either as a document or a in [github](https://docs.github.com/en/issues/planning-and-tracking-with-projects/customizing-views-in-your-project/changing-the-layout-of-a-view#about-the-roadmap-layout)

scfw/cli.py
```diff
@@ -36,16 +36,22 @@ def _add_configure_cli(parser: ArgumentParser):
         help="Remove all Supply-Chain Firewall-managed configuration"
     )
 
+    parser.add_argument(
+        "--alias-npm",
+        action="store_true",
+        help="Add shell aliases to always run npm commands through Supply-Chain Firewall"
+    )
+
```
sobregosodd, line 44: foreach SUPPORTED_PACKAGE_MANAGERS here?

scfw/commands/poetry_command.py
```diff
@@ -0,0 +1,131 @@
...
+
+        def line_to_install_target(line: str) -> Optional[InstallTarget]:
+            # All supported versions adhere to this format
+            match = re.search(r"(Installing|Updating|Downgrading) (.*) \((.*)\)", line.strip())
+            if match and "Skipped" not in line:
+                return InstallTarget(self.ecosystem(), match.group(2), get_target_version(match.group(3)))
+            return None
+
+        # For now, automatically allow all non-`add` commands
+        if "add" not in self._command:
```
sobregosodd, line 116: - [ ] AFAIK _command includes the command line which contains a package name, if so, this will run with execution of `poetry show some-package-add-name` (for example)

scfw/configure/interactive.py
```diff
@@ -29,14 +29,19 @@ def get_answers() -> dict:
     has_dd_api_key = os.getenv(DD_API_KEY_VAR) is not None
 
     questions = [
```
sobregosodd, line 31: for?

tests/commands/test_poetry.py
```diff
@@ -0,0 +1,243 @@
...
+    modify the local installation state.
+    """
+    subprocess.run(command, check=True, cwd=new_poetry_project)
+    assert poetry_show(new_poetry_project) == init_poetry_state
+
+
+@pytest.mark.parametrize(
+        "command",
+        [
+            ["poetry", "add", "!a_nonexistent_p@ckage_name"],
```
sobregosodd, line 136: can we add the poetry show example? to make sure it does not runs

tests/commands/test_poetry_command.py
```diff
@@ -0,0 +1,83 @@
...
+    poetry_project,
+    target_spec,
+    target_name,
+    target_version,
+):
+    """
+    Tests that `PoetryCommand.would_install()` correctly resolves installation
+    targets for a variety of target specfications without installing anything.
+    """
+    if poetry_project == "new_poetry_project":
```
sobregosodd, line 54: cant you parametrize the function instead of its name?

sobregosodd (changes requested):
just one concern

scfw/cli.py
ikretz: This is a good idea.  There may be other opportunities elsewhere in the code to reuse `SUPPORTED_PACKAGE_MANAGERS` to simply configuration things.  Let's do this in the next one.

README.md
ikretz: Good point.  Let's change the column header to "Inspected subcommands".  Then I'll clarify lower down that all commands can run.

README.md
ikretz: I will delete this line.  A public roadmap is a good thing, let's link it when we have it.

scfw/commands/poetry_command.py
ikretz: Yes and no, but I see what you're saying.  `self._command` is the list of tokens passed on the command line, so `["poetry", "show", "some-package-add-name"]` for your example.  So that won't cause a problem with `"add" in self._command` in this case.  However, if the user passes a command with an argument `"add"`, that will be a false positive.

The approach we've taken with the other `PackageManagerCommands` is that we can't have any false negatives for the subcommands we support.  However, to rule out false positives, we would have to write a full parser for each package manager's CLI.  That may be something we want to do eventually, but for now, we take this very simple approach that still ensures no false negatives.

tests/commands/test_poetry.py
ikretz: See previous comment.

tests/commands/test_poetry_command.py
ikretz: You can't pass the fixtures as arguments in the `parametrize` decorator, if you do, they get interpreted as the functions themselves instead of their return values.  At least, I wasn't able to get it to work.  Apparently this trick is an accepted one.

scfw/configure/interactive.py
ikretz: Yeah, here's another spot where we can reuse `SUPPORTED_PACKAGE_MANAGERS`.  Let's do this in the next one.  I think there will be other such opportunities in here.

scfw/commands/poetry_command.py
ikretz: And in the case of false positives, the user can temporarily `unalias poetry` if they need to.  It's not ideal but there is a simple obvious workaround.

README.md
ikretz: Changed in [efa59d3](https://github.com/DataDog/supply-chain-firewall/pull/74/commits/efa59d3e5ba77b40e9b7c7771436909823fecfa9)

README.md
ikretz: Changed in [efa59d3](https://github.com/DataDog/supply-chain-firewall/pull/74/commits/efa59d3e5ba77b40e9b7c7771436909823fecfa9)

sobregosodd (approved):

scfw/commands/poetry_command.py
sobregosodd: this was my only concern

PR #78, opened 2025-05-02
Add support for Poetry install

ikretz:
This PR adds support for `poetry install` commands to SCFW.

It also performs some light refactoring and updates dependencies to their latest compatible versions.


ikretz:
Restarting in [ikretz/poetry-install-2](https://github.com/DataDog/supply-chain-firewall/tree/ikretz/poetry-install-2) to delay refactoring around the use of `SUPPORTED_PACKAGE_MANAGERS` to a later, broader refactoring of the way command-line arguments are handled to be more structured.

PR #79, opened 2025-05-04
Add support for poetry install

ikretz:
This PR adds support for the `poetry install` subcommand along with related tests.

In doing so, it **temporarily** drops support for Poetry before 1.7.0.  This is because Poetry before 1.7.0 does not include the package being installed among the installation targets reported in its dry-run output.  There is a workaround we could pursue to account for this down the road once we have fully supported all "installish" Poetry subcomands in this reduced version range.

It also updates dependencies to their latest compatible versions.


scfw/commands/poetry_command.py
```diff
@@ -107,13 +107,12 @@ def get_target_version(version_spec: str) -> str:
...
-            match = re.search(r"(Installing|Updating|Downgrading) (.*) \((.*)\)", line.strip())
-            if match and "Skipped" not in line:
+            pattern = r"(Installing|Updating|Downgrading) (?:the current project: )?(.*) \((.*)\)"
+            if "Skipped" not in line and (match := re.search(pattern, line.strip())):
                 return InstallTarget(self.ecosystem(), match.group(2), get_target_version(match.group(3)))
             return None
 
-        # For now, automatically allow all non-`add` commands
-        if "add" not in self._command:
+        if not any(subcommand in self._command for subcommand in {"add", "install"}):
```
sobregosodd, line 115: instead of using the subcommands like this, maybe we should introduce the `subcommands` class attribute in `PackageManagerCommand`, which each package manager should override.
We can also use this programatically to build the docs (`inspected subcommands` column)

tesnim5hamdouni (approved):

scfw/commands/poetry_command.py
ikretz: Ooh, nice, I like it.  I'll make a ticket for this.

PR #83, opened 2025-05-06
[SECRES-3058] Add support for poetry update

ikretz:
This PR adds support for the `poetry update` subcommand including related tests.

All "installish" Poetry subcommands are now supported.

### DIFF
--- a/tests/commands/test_poetry.py
+++ b/tests/commands/test_poetry.py
@@ -99,6 +99,21 @@ def poetry_project_target_previous_lock_latest():
     tempdir.cleanup()
 
 
+@pytest.fixture
+def poetry_project_lock_latest():
+    """
+    Initialize a Poetry project where the latest version of `TARGET` is an as-yet
+    uninstalled dependency.
+    """
+    tempdir = TemporaryDirectory()
+    _init_poetry_project(tempdir.name, TEST_PROJECT_NAME)
+    subprocess.run(["poetry", "add", "--lock", f"{TARGET}=={TARGET_LATEST}"], check=True, cwd=tempdir.name)
+
+    yield tempdir.name
+
+    tempdir.cleanup()
+
+
 def test_poetry_version_output():
     """
     Test that `poetry --version` has the required format and parses correctly.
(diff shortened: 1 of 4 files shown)


tests/commands/test_poetry.py
```diff
@@ -183,6 +198,29 @@ def test_poetry_sync_no_change(new_poetry_project):
...
+        ["poetry", "update", "-V"],
+        ["poetry", "--version", "update"],
+        ["poetry", "update", "--version"],
+        ["poetry", "-h", "update"],
+        ["poetry", "update", "-h"],
+        ["poetry", "--help", "update"],
+        ["poetry", "update", "--help"],
+        ["poetry", "--dry-run", "update"],
+        ["poetry", "update", "--dry-run"],
+    ]
```
tesnim5hamdouni, line 217: you may want to use @pytest.mark.parametrize

tests/commands/test_poetry.py
```diff
@@ -273,6 +311,35 @@ def test_poetry_sync_error_no_change(new_poetry_project):
...
+        ["poetry", "update", TARGET, "--only"],
+        ["poetry", "update", "-P"],
+        ["poetry", "update", TARGET, "-P"],
+        ["poetry", "update", "--project"],
+        ["poetry", "update", TARGET, "--project"],
+        ["poetry", "update", "-C"],
+        ["poetry", "update", TARGET, "-C"],
+        ["poetry", "update", "--directory"],
+        ["poetry", "update", TARGET, "--directory"],
+    ]
```
tesnim5hamdouni, line 336: and here as well

tesnim5hamdouni (approved):
how does poetry deal with corrupted lock files?

tests/commands/test_poetry.py
ikretz: I was using `parametrize` before but moved away from it because it was causing the tests to run really slowly.  Like 20 minutes for each version of Poetry in CI.

The reason for this was that `pytest` reinitializes the fixtures once for each `parametrize` test case instead of once for each test.  And Poetry is really slow, so it was really adding up.

Doing it this way strikes a good balance, I think.  Now the tests only take like 7 minutes to complete 😄

PR #88, opened 2025-05-26
[SECRES-3126] Add local auditing of installed packages

ikretz:
This PR adds to SCFW the ability to verify installed packages in local environments:

```shell
$ scfw audit --help
usage: scfw audit [-h] [--executable PATH] {npm,pip,poetry}

Audit installed packages using Supply-Chain Firewall's verifiers.

positional arguments:
  {npm,pip,poetry}   The package manager whose installed packages should be verified

options:
  -h, --help         show this help message and exit
  --executable PATH  Package manager executable to use for running commands (default: environmentally determined)
```

```shell
$ scfw audit npm
No issues found.

$ scfw audit pip
Package pip-23.0.1:
  - An OSV.dev advisory exists for package pip-23.0.1:
      * [Medium] https://osv.dev/vulnerability/GHSA-mq26-g339-26xf
  - An OSV.dev advisory exists for package pip-23.0.1:
      * [Low] https://osv.dev/vulnerability/PYSEC-2023-228
Package setuptools-65.5.0:
  - An OSV.dev advisory exists for package setuptools-65.5.0:
      * [High] https://osv.dev/vulnerability/GHSA-cx63-2mw6-8hw5
  - An OSV.dev advisory exists for package setuptools-65.5.0:
      * [High] https://osv.dev/vulnerability/GHSA-5rjg-fvgr-3xxf
  - An OSV.dev advisory exists for package setuptools-65.5.0:
      * [High] https://osv.dev/vulnerability/GHSA-r9hx-vwmv-q579
  - An OSV.dev advisory exists for package setuptools-65.5.0:
      * https://osv.dev/vulnerability/PYSEC-2022-43012
```

The underlying package managers are used to determine installed packages.  These are then run through SCFW's verifiers.  Findings are collected into a single report, organized by package and sorted by severity.  If there are no findings, a message to this effect is printed.

For `npm` and `poetry`, SCFW assumes the project of interest is in the current working directory.

Currently, audit actions are not logged to Datadog.  We can add this in a subsequent PR before releasing this feature.

Other changes include:
* Refactoring the `run` subcommand CLI slightly to make parsing simpler
* Optimization: don't prompt the user for confirmation when `--dry-run` option was passed 
* Adding tests and documentation for the new feature
* Replace `would_install` with `resolve_install_targets` in package manager test function names

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,9 +114,9 @@ Package setuptools-65.5.0:
       * https://osv.dev/vulnerability/PYSEC-2022-43012
 ```
 
-Supply-Chain Firewall audits all installed packages visible to the package manager in the invoking environment.  The user may specify the package manager executable they wish to use on the command line.
+Supply-Chain Firewall audits all installed packages visible to the package manager in the invoking environment.  The user may specify the package manager executable they wish to use on the command line.  For `npm` and `poetry` audits, Supply-Chain Firewall assumes the project of interest is in the current working directory.
 
-For `npm` and `poetry` audits, Supply-Chain Firewall assumes the project of interest is in the current working directory.  Currently, `npm` audits do not take globally installed packages into consideration.
+Currently, `npm` audits do not take globally installed packages into consideration.  To audit a globally installed `npm` package, first `cd` into the package's directory (inside the global `node_modules/`) and perform a local audit there.
 
 ## Datadog Log Management integration
 


README.md
```diff
@@ -87,6 +89,31 @@ $ scfw run poetry add git+https://github.com/DataDog/guarddog
 
 For `pip install` commands, packages will be installed in the same environment (virtual or global) in which the command was run.
 
+### Audit installed packages
+
+Supply-Chain Firewall can also use its verifiers to audit installed packages:
```
sobregosodd, line 94: i'm not sure this is clear enough. IMO, it should:
- if the user provides a package name it should narrow down the validation only to that one. 
- if the user provides no package as param it should perform the audit to the entire environ.

but doing the audit to the entire environ is not as straight forward as every lang is different

don't you agree?

scfw/package_managers/__init__.py
```diff
@@ -10,38 +10,38 @@
 from scfw.package_managers.pip import Pip
 from scfw.package_managers.poetry import Poetry
 
-SUPPORTED_PACKAGE_MANAGERS = {
+SUPPORTED_PACKAGE_MANAGERS = [
```
sobregosodd, line 13: curious, why the change here?

scfw/package_managers/npm.py
```diff
@@ -163,6 +164,43 @@ def str_to_package(s: str) -> Package:
...
+
+            for name, package_data in dependencies.items():
+                if (package_dependencies := package_data.get("dependencies")):
+                    packages |= dependencies_to_packages(package_dependencies)
+                packages.add(Package(ECOSYSTEM.Npm, name, package_data["version"]))
+
+            return packages
+
+        try:
+            npm_list_command = self._normalize_command(["npm", "list", "--all", "--json"])
```
sobregosodd, line 190: mind that this will check packages in the CWD.
what happens with the global installed packages?

scfw/package_managers/__init__.py
ikretz: Ahh, it's just a little thing: it's so I can use `SUPPORTED_PACKAGE_MANAGERS` as the set of argument choices for the `scfw audit` subcommand (`cli.py:35`), specifically so that the choices are printed in a definite order in the help message.  If I use a set, they could get printed in any order (same if I do `list(SUPPORTED_PACKAGE_MANAGERS)` at the point of use).

README.md
ikretz: I had not considered a use-case where you want to audit only a given installed package.  I think this could be nice but let's treat it as an enhancement and save it for the next PR.

As for auditing the entire environment, SCFW uses a very simple approach: ask whatever package manager is active in the current environment what it thinks the installed packages are.  This approach is the correct one because, with the exception of `npm` (more on that later), the packages that are installed and available in the current environment are exactly those reported by this environment-specific package manager.

Breaking it down for the well-behaved package managers:

* `poetry`: Has no concept of global versus local environments.  Everything is oriented around the idea of a project, defined by a `pyproject.toml`/`poetry.lock` pair.  Running `poetry show` reflects what has been installed for a given project
* `pip`: Maintains a single, global environment of installed packages.  Can create local, isolated install sites via the use of `venv`/`virualenv`.  Running `pip list` via the locally available `python` executable accurately reflects what packages are available in the current environment (virtual or otherwise)

On the other hand, in `npm` world, there are always two environments that coexist with one another: a local environment containing packages installed for a particular project as well as a global environment of packages not associated with any project.  By default, all package operations take place in the local environment, which conforms to SCFW's model.  Thus, when the user runs `scfw audit npm`, only the local environment is audited.  Currently, the global environment is ignored (more on this in the next thread).

All this to say, auditing the entire environment is straightforward (modulo this irregularity with `npm`).  I will add some more detail to this section of the README to clarify what the audit mode is for and what the current limitations are.

scfw/package_managers/npm.py
ikretz: We're currently ignoring the global environment for `audit npm`.  The previous thread has the relevant context.

Long-term, I think the right move is to add a `--global` switch to the `audit npm` subcommand so the user can specify which environment they want to audit.  Alternately, since packages in the global environment are available in every local environment, we could always include these in the audit.  Let me look into this a little bit.

README.md
ikretz: Added in [804fe8b](https://github.com/DataDog/supply-chain-firewall/pull/88/commits/804fe8ba8c0e747346ba0231e3c398b841cf1a02)

scfw/package_managers/npm.py
ikretz: In the interest of keeping the PR small(ish) and focused, I've decided for this initial release to always include the globally installed packages in every `npm` audit (added in [547b13a](https://github.com/DataDog/supply-chain-firewall/pull/88/commits/547b13aee7c30e606122c1f3ff57fe7ba6f50a72)).

We can add the `--global` option to allow the user to specify where they want to audit in the next one.

ikretz: Long story short (see gnarly commit history below), I reverted this change: globally installed packages are now (as before) ignored during `npm` audits.

I was forced to do this because it turns out that `npm@10.6.0` is in some kind of weird dependency hell where for three of its dependencies, the versions that get installed (on a clean Node.js install) are incompatible with what is specified in its `package.json` file.  This causes the `npm list` command we are running behind the scenes to error out, and as a result, we break support for `npm@10.6.0` if we always audit the globals.

I also had the realization that if the user wants to audit the globally installed packages, they can just `cd` into the package's directory under `lib/node_modules/` and perform a local audit there.  So there is still a way for the user to audit the globals with SCFW.

PR #103, opened 2025-07-01
[SECRES-3222] [SECRES-3272] Add Datadog logging for local audits

ikretz:
This PR adds Datadog logging for the `scfw audit` command.

Unlike package manager operations, which occur in the course of normal development workflows, an audit is a special-purpose operation, one that the user can only trigger by invoking `scfw` directly (as opposed to indirectly via shell aliases) and usually occurring in an incident response-like context.  As such, we consider that there is always in interest in logging the results of an audit, and so **the configured `SCFW_LOG_LEVEL` settings are ignored when it comes to logging audits**.  More precisely, we consider that these settings pertain only to the firewall operations of inspecting package manager commands.

Other changes include:
* Log the command-line name of the package manager in firewall action logs
* Add unit tests for the malicious packages verifier's caching logic
* Fallback on cached manifests on any error, not just `requests.HTTPError` (not general enough)
* Reduce error message verbosity when the OSV.dev API fails
* Version number bump in preparation for next release

### DIFF
--- a/scfw/firewall.py
+++ b/scfw/firewall.py
@@ -48,7 +48,7 @@ def run_firewall(args: Namespace) -> int:
             reports = verifiers.verify_packages(targets)
 
             if (critical_report := reports.get(FindingSeverity.CRITICAL)):
-                loggers.log(
+                loggers.log_firewall_action(
                     package_manager.ecosystem(),
                     package_manager.executable(),
                     args.command,
@@ -65,7 +65,7 @@ def run_firewall(args: Namespace) -> int:
                 warned = True
 
                 if not args.dry_run and not inquirer.confirm("Proceed with installation?", default=False):
-                    loggers.log(
+                    loggers.log_firewall_action(
                         package_manager.ecosystem(),
                         package_manager.executable(),
                         args.command,
@@ -80,7 +80,7 @@ def run_firewall(args: Namespace) -> int:
             _log.info("Firewall dry-run mode enabled: command will not be run")
             print("Dry-run: exiting without running command.")
         else:
-            loggers.log(
+            loggers.log_firewall_action(
                 package_manager.ecosystem(),
                 package_manager.executable(),
                 args.command,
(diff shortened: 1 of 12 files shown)


scfw/firewall.py
```diff
@@ -48,8 +48,9 @@ def run_firewall(args: Namespace) -> int:
             reports = verifiers.verify_packages(targets)
 
             if (critical_report := reports.get(FindingSeverity.CRITICAL)):
```
tesnim5hamdouni, line 50: Maybe consider refactoring these calls.
Since they almost have the same arguments, you can define a list of each case and then call the loggers
```
if (critical_report := reports.get(FindingSeverity.CRITICAL)):
    packages = list(critical_report.packages())
elif (warning_report := reports.get(FindingSeverity.WARNING)):
    packages = list(warning_report.packages())
else:
    packages = targets

loggers.log_firewall_action(
    package_manager.ecosystem(),
    package_manager.name(),
    package_manager.executable(),
    args.command,
    packages,
)
```

tesnim5hamdouni (approved):
Waiting for Sebastian's review

scfw/firewall.py
ikretz: We can have an auxiliary function that will cut down on some of the repetitious elements (e.g., all the `package_manager.` stuff), but I don't want to refactor these down too much because they act as a form of documentation for what happens in each case.

ikretz: I tried the auxiliary function and the extra indirection doesn't really save us any verbosity.  Instead, I think there's an opportunity to refactor `FirewallLoggers` itself, and I think I prefer to go this way.  However, this would be a separate ticket/PR.  I'll schedule it.

scfw/logger.py
```diff
@@ -71,24 +73,26 @@ def from_string(cls, s: str) -> Self:
 
 class FirewallLogger(metaclass=ABCMeta):
     """
-    An interface for passing information about a completed firewall run to
+    An interface for passing information about runs of Supply-Chain Firewall to
     client loggers.
     """
     @abstractmethod
-    def log(
+    def log_firewall_action(
```
sobregosodd, line 80: `log_action` was shorter IMO

sobregosodd (approved):

scfw/logger.py
ikretz: It's to match the `FirewallAction` class name, but we can make it less verbose in the next one.

PR #106, opened 2025-07-07
[SECRES-3270] Allow verifiers to support only certain ecosystems

ikretz:
This PR allows package verifiers to support verification for only a subset of ecosystems.  Previously, each verifier was required to support all ecosystems.

Both default verifiers have been modified to raise a `WARNING` finding when an attempt is made to verify a package in an unsupported ecosystem, although both verifiers support all ecosystems for now.  Moreover, the `FirewallVerifiers` init routine has been modified to raise a `RuntimeError` if it fails to discover at least one verifier that supports the given ecosystem.  This is all to prevent null checks.

Other changes include:
* Light refactoring in `DatadogMaliciousPackagesVerifier` to take advantage of new `supported_ecosystems` method
* Updating `DatadogMaliciousPackagesVerifier` tests in light of refactoring


scfw/verifiers/osv_verifier/__init__.py
```diff
@@ -59,8 +70,18 @@ def finding(osv: OsvAdvisory) -> str:
...
+        def failure_message() -> str:
+            return (
+                f"Failed to verify package {package} via the OSV.dev API.\n"
+                f"Before proceeding, please check the OSV.dev website for advisories related to this package.\n"
+                f"DO NOT PROCEED if the package has advisories with a MAL ID: it is very likely malicious.\n"
+                f"  * {_OSV_DEV_LIST_URL_PREFIX}?q={package.name}&ecosystem={str(package.ecosystem)}"
+            )
 
+        if package.ecosystem not in self.supported_ecosystems():
+            return [(FindingSeverity.WARNING, f"Package ecosystem {package.ecosystem} is not supported")]
```
sobregosodd, line 82: Ain't better to do super here? instead of repeating

ikretz: Maybe, although I like that there's no connections between the two verifier implementations right now.

There is probably a generic framework we could provide for a verifier implementation but it would be a significant change, and I would prefer to do it all at once instead of bit-by-bit.  I'll make a ticket to look into it.

PR #117, opened 2025-08-04
[SECRES-3391] Handle compromised libs in malicious packages dataset verifier

ikretz:
This PR enhances SCFW's package verifier for Datadog Security Research's malicious packages [dataset](https://github.com/DataDog/malicious-software-packages-dataset) to honor version numbers for benign installation targets that have previously suffered compromises.  Previously, this verifier ignored version numbers and performed verification based only on package names, which would have effectively prevented users of such packages from installing even patched versions.

Other changes:
* Enhance the DD verifier tests to include both malicious intent and compromised lib packages

### DIFF
--- a/tests/verifiers/test_dd_verifier.py
+++ b/tests/verifiers/test_dd_verifier.py
@@ -8,7 +8,6 @@ from pathlib import Path
 import pytest
 from tempfile import TemporaryDirectory
 
-from scfw.configure import SCFW_HOME_VAR
 from scfw.ecosystem import ECOSYSTEM
 from scfw.package import Package
 from scfw.verifier import FindingSeverity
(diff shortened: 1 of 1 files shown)


scfw/verifiers/dd_verifier/__init__.py
```diff
@@ -68,13 +68,14 @@ def verify(self, package: Package) -> list[tuple[FindingSeverity, str]]:
...
-        if package.name in manifest:
+        if (
+            package.name in manifest
+            and (not manifest[package.name] or package.version in manifest[package.name])
+        ):
             return [
                 (
                     FindingSeverity.CRITICAL,
-                    f"Datadog Security Research has determined that package {package.name} is malicious"
+                    f"Datadog Security Research has determined that package {package} is malicious"
```
tesnim5hamdouni, line 78: I suggest going with a more user friendly dispaly of the package in question

sobregosodd (approved):

tesnim5hamdouni (changes requested):

scfw/verifiers/dd_verifier/__init__.py
ikretz: This string ensures the package data is formatted according to ecosystem-specific conventions,`name-version` (PyPI) or `name@version` (npm).

Can you please clarify what you mean by "user-friendly"?

tesnim5hamdouni: that works then

PR #143, opened 2025-09-23
[SECRES-3532] Delay check for supported package manager version based on need

ikretz:
This PR optimizes the behavior of SCFW's `PackageManager` implementations regarding supported version checking.

Currently, SCFW checks that the version of the underlying package manager is supported at `PackageManager` initialization: instantiating the `PackageManager` is impossible unless this version check passes.  While this approach is sound from a security standpoint, it is overly conservative: having a supported version is only needed when SCFW knows it has work to do, that is, when it is `run`ing  an "installish" package manager command or performing an `audit`.  Otherwise, enforcing version compliance ends up blocking by default commands for which SCFW is irrelevant.  Thus, this PR delays the version checks as long as possible until we know SCFW has work to do.

Other changes include:
* Adding matrix tests for the latest version of npm, pip and Poetry

### DIFF
--- a/scfw/package_managers/npm.py
+++ b/scfw/package_managers/npm.py
@@ -36,14 +36,6 @@ class Npm(PackageManager):
         Raises:
             RuntimeError: A valid executable could not be resolved.
         """
-        def get_npm_version(executable: str) -> Optional[Version]:
-            try:
-                # All supported versions adhere to this format
-                npm_version = subprocess.run([executable, "--version"], check=True, text=True, capture_output=True)
-                return version_parse(npm_version.stdout.strip())
-            except InvalidVersion:
-                return None
-
         executable = executable if executable else shutil.which(self.name())
         if not executable:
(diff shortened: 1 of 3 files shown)


scfw/package_managers/npm.py
```diff
@@ -189,6 +189,8 @@ def dependencies_to_packages(dependencies: dict[str, dict]) -> set[Package]:
 
             return packages
 
+        self._check_version()
```
sobregosodd, line 185: you can make _version into a singleton pattern, thus you'll be achieving lazy evaluation, and you don't have to worry about when to check the version

ikretz: Moved the "get version" step into the "check version" step in [5340f20](https://github.com/DataDog/supply-chain-firewall/pull/143/commits/5340f20c7ac15e1ad177cc5f3eb731cb3bbc49be)

PR #145, opened 2025-09-24
[SECRES-3536] Make dry-run exit code reflect findings status

### LINKED ISSUE #139: CI usage of scfw run --dry-run pip install
Great tool!

We're aiming to use `scfw run --dry-run pip install -r requirements.txt` to verify a pip requirements file in a CI pipeline, without actually installing.
We'd like to have it return a non-zero exit code if any malware or vulnerability is found.

Is there any way to achieve this without postprocessing the scfw output? I've already tried `--block-on-warning` and `--error-on-block` but neither seems to achieve my use case (exit code is 0 even if vulnerabilities are found)

Thanks in advance for any guidance!

ikretz:
This PR makes it so SCFW's exit code when the `--dry-run` option is given reflects whether or not there were findings of any severity, with 0 indicating no findings.  Previously, the exit code for dry-runs was always 0.

This PR also fixes a bug with respect to `--dry-run` and `CRITICAL` findings: if `--dry-run` is given, SCFW no longer sends a `BLOCK` log, as it was previously doing.  The new behavior is correct because dry-runs are not `BLOCK` actions.

In order to achieve these two goals, the PR refactors `run_firewall()` slightly so that it no longer uses a separate boolean variable to track whether it warned the user.

Closes #139

### DIFF
--- a/scfw/firewall.py
+++ b/scfw/firewall.py
@@ -44,8 +44,10 @@ def run_firewall(args: Namespace) -> int:
             _log.info(f"Using package verifiers: [{', '.join(verifiers.names())}]")
 
             reports = verifiers.verify_packages(targets)
+            critical_report = reports.get(FindingSeverity.CRITICAL)
+            warning_report = reports.get(FindingSeverity.WARNING)
 
-            if (critical_report := reports.get(FindingSeverity.CRITICAL)) and not args.dry_run:
+            if not args.dry_run and critical_report:
                 loggers.log_firewall_action(
                     package_manager.ecosystem(),
                     package_manager.name(),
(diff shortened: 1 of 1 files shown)


sobregosodd:
> Previously, we were only printing CRITICAL findings or WARNING findings, not both
@ikretz  don't we have a log level to rely on instead of hardcoding this decision?

ikretz:
> > Previously, we were only printing CRITICAL findings or WARNING findings, not both
> 
> @ikretz don't we have a log level to rely on instead of hardcoding this decision?

We have configurable log levels for the local logging output (`DEBUG`, `INFO`, etc.) and for what gets sent to Datadog (via `SCFW_DD_LOG_LEVEL`).  For the dry-run output, we were supposed to just be printing everything we found, so this part is actually a bug fix.

PR #147, opened 2025-09-25
Require stricter opt-in for Datadog API logging

ikretz:
This PR makes it so users must opt-in to Datadog API logging by setting a new, SCFW-specific environment variable, `SCFW_DD_API_LOGGER_ENABLED`.

Previously, the mere presence of `DD_API_KEY` in the environment was taken as this opt-in "signal."  This turns out to be a bad signal in practice: API keys for non-US1 orgs need additional configuration (e.g., `DD_SITE`), causing errors if not present.  In such cases, SCFW was failing at the logging step for allowed commands and thus never running the package manager command, denying service to these users.

This PR also makes it so SCFW no longer handles Datadog API keys, either directly or indirectly:
* During configuration, the tool no longer prompts for or accepts Datadog API keys 
* During logger initialization, the tool no longer checks for `DD_API_KEY` in the environment

This choice was made because SCFW would have trouble anticipating all of the particulars of configuring the environment for API key use, making the utility of the `configure` subcommand dubious for this use-case.  Users who wish to opt into Datadog API logging are better able to manage this configuration for themselves.


ikretz:
Superseded by [45bd998](https://github.com/DataDog/supply-chain-firewall/commit/45bd998c9fd28bbf9c732dbb41b56344ae09bda7)

PR #150, opened 2025-09-29
[SECRES-3783] Warn on logger errors instead of exiting early

ikretz:
This PR changes the behavior of SCFW so that we no longer exit early on logger error.  Now, we continue running and simply warn on logger errors.  This fixes an issue where early exits on logger errors ended up denying service in the case of allowed commands (since logging occurs before command execution).

Other changes:
* Improving error messages in the Datadog loggers

### DIFF
--- a/scfw/loggers/__init__.py
+++ b/scfw/loggers/__init__.py
@@ -88,4 +88,4 @@ class FirewallLoggers(FirewallLogger):
             try:
                 logger.log_audit(ecosystem, package_manager, executable, reports)
             except Exception as e:
-                _log.warning(f"Failed to log audit action: {e}")
+                _log.warning(f"Failed to log audit: {e}")


scfw/loggers/__init__.py
```diff
@@ -82,4 +85,7 @@ def log_audit(
         Log the results of an audit to all client loggers.
         """
         for logger in self._loggers:
-            logger.log_audit(ecosystem, package_manager, executable, reports)
+            try:
+                logger.log_audit(ecosystem, package_manager, executable, reports)
+            except Exception as e:
+                _log.warning(f"Failed to log audit: {e}")
```
sobregosodd, line 91: Log warning? ain't it an error

ikretz: It is an error, but so far we've been reserving `log.error()` for unrecoverable errors, the ones that cause an immediate exit.  Since SCFW can tolerate a log failure, I categorized these as warnings.

PR #158, opened 2025-10-07
[SECRES-3850] Handle non-interactive shells gracefully

ikretz:
This PR adds a check during `scfw run` execution for whether SCFW is running in a non-interactive shell.  This is relevant in the case where the user must provide interactive input to confirm a command with only warning-level findings.  Instead of erroring out, SCFW now defaults to blocking in this situation.  Users can change this default behavior by adding `--allow-on-warning` to their SCFW command-line.  

In some cases, however, users are unable or unwilling to update their command lines.  For this reason, this PR also allows users to configure this auto-allow/auto-block behavior by setting environment variables `SCFW_ALLOW_ON_WARNING` or `SCFW_BLOCK_ON_WARNING`.

The command-line options and environment variables determine the warning-level action as follows:
1. If `--block-on-warning` or `SCFW_BLOCK_ON_WARNING` is set, `BLOCK`
2. Else if `--allow-on-warning` or `SCFW_ALLOW_ON_WARNING` is set, `ALLOW`
3. Else if in a non-interactive shell, `BLOCK`
4. Else, ask the user for confirmation and act accordingly

Other changes include:
* Adding more precise error checking and logs while configuring the Datadog Agent
* Move the file containing configuration-related constants to the top-level directory


scfw/firewall.py
```diff
@@ -128,3 +128,32 @@ def run_firewall(args: Namespace) -> int:
...
+        cli_block_choice:
+            A `bool` indicating whether the user selected `--block-on-warning` on the command-line.
+
+    Returns:
+        The `FirewallAction` that should be taken based on the user's configured choices or, if
+        no choice has been made, on the user's runtime (interactive) decision.
+    """
+    if cli_block_choice or os.getenv(BLOCK_ON_WARNING_VAR):
+        return FirewallAction.BLOCK
+    if cli_allow_choice or os.getenv(ALLOW_ON_WARNING_VAR):
```
tesnim5hamdouni: 1. This supposes that `ALLOW_ON_WARNING_VAR` and `BLOCK_ON_WARNING_VAR` are mutually exclusive which is not the case. 
As a user, I'll have to set the `ALLOW...` AND unset the `BLOCK..` which is not necessarily intuitive or at least requires more explanation in the readme.
2. Another caveat is that if I forget to unset the env variables my cli `--allow` command will be useless. Why not consider this order: check for cli commands first, if none are set default to env variables with priority to blocking, else if interactive shell block, else ask user
3. why not define a single env variable if the intended behavior is mutual exclusion? (I suppose either you block or allow, in which case a binary state is better )

ikretz: Agreed, why don't we do this:
* One env var with values `{ALLOW, BLOCK}`, which will solve a lot of these problems
* The CLI options can override the env variable as you say

ikretz: Resolved in [5d05381](https://github.com/DataDog/supply-chain-firewall/pull/158/commits/5d0538197cdde688fe749e885e9689bac55c5fd9)

PR #162, opened 2025-10-15
[SECRES-3967] Add Go support

ikretz:
This PR adds support for the Go ecosystem and the `go` command line to Supply-Chain Firewall.

A huge thanks to @SirGFM for developing this feature and contributing it back to SCFW!

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -49,34 +49,6 @@ jobs:
       - name: Test firewall configure script
         run: make test-configure
 
-  go-integration:
-    runs-on: ubuntu-latest
-    strategy:
-      fail-fast: false
-      matrix:
-        go-version: ["1.17", "1.18", "1.19", "1.20", "1.21", "1.22", "1.23", "1.24", "1.25"]
-    steps:
-      - uses: actions/checkout@v4
-      - name: Set up Python 3.10
-        uses: actions/setup-python@v5
-        with:
-          python-version: "3.10"
-      - name: Install dependencies
-        run: |
-          python -m pip install --upgrade pip
-          pip install -r requirements.txt
-          pip install -r requirements-dev.txt
-      - name: Install Supply-Chain Firewall
-        run: pip install .
-      - name: Install Go toolchain
-        uses: actions/setup-go@v6
-        with:
-          go-version: ${{ matrix.go-version }}
-      - name: Test firewall Go integration
-        run: |
-          make test-go
-          make test-go-class
-
   python-executable:
     runs-on: ubuntu-latest
     steps:


scfw/package_managers/go/temp_env.py
```diff
@@ -0,0 +1,160 @@
...
+        # Go searches all directories listed in GOPATH to find source code but
+        # new packages are always downloaded into the first directory in the list
+        self._gopath = f"{self._go_dir}:{get_gopath(executable)}"
+
+        # Override Go-specific environment variables so that all package
+        # downloading and caching occurs within the disposable `TempGoEnvironment`
+        self._env = os.environ.copy()
+        self._env['GOCACHE'] = str(self._cache_dir)
+        self._env['GOMODCACHE'] = str(self._mod_cache_dir)
+        self._env['GOPATH'] = self._gopath
```
SirGFM, line 87: I misunderstood how go behaves... Before go module, it was possible to specify multiple directories in `GOPATH` and go would install new packages in the first path while using any others as a cache for reading. However, since then, this has changed and now only `GOCACHE` and `GOMODCACHE` are used to store downloaded packages.

Thus, since a clean cache is generated for each new command, this ends up making packages get re-downloaded every time scfw runs (which can be quite slow).

To solve this, we introduce a global cache in our go port that maintains `GOCACHE` and `GOMODCACHE` until a critical finding is detected, at which point this entire cache is purged. This way, packages are downloaded only once, speeding up the execution of scfw by a lot.

Here's how we did it in the go port: https://github.com/New-Horizons-Team/supply-chain-firewall/compare/8e3cccc9e5339e437df2db5aa6876e352a3b524b...New-Horizons-Team:supply-chain-firewall:11b275339028222f91a5c3815942f4a9daa596f1

scfw/package_managers/go/__init__.py
```diff
@@ -0,0 +1,243 @@
...
+            local_packages, remote_packages = set(), set()
+
+            for param in inspected_subcommand[1:]:
+                if param.startswith("-"):
+                    continue
+                # TODO(ikretz): Handle packages with envvars?
+                if Path(param).exists() or len(param.split("/")) == 1:
+                    local_packages.add(param)
+                else:
+                    remote_packages.add(param)
```
SirGFM, line 149: There are quite a few go commands that may take arguments, which we hadn't originally ignored. Thus, running something like `go build -o <target> .` would end up considering `<target>` as a package to be inspected.

Essentially, `build`, `clean`, `get`, `install`, `list`, and `run` accept a list of common "build flags", while `build`, `generate`, and `run` each have specific flags.

We've adjusted this in our go port over these commits: https://github.com/New-Horizons-Team/supply-chain-firewall/compare/058426bcdec07127a5753fb377ed209896c2e854...New-Horizons-Team:supply-chain-firewall:8e3cccc9e5339e437df2db5aa6876e352a3b524b

SirGFM (commented):
Sorry it took so long to get back about this.

These are the points that I had mentioned that we found and fixed in the go port.

PR #174, opened 2025-11-21
[SECRES-4057] Update PyPI publish workflow to use Trusted Publisher

ikretz:
This PR updates the `pypi-release.yml` GitHub Action to use Datadog's Trusted Publisher.

### DIFF
--- a/.github/workflows/pypi-release.yml
+++ b/.github/workflows/pypi-release.yml
@@ -11,6 +11,9 @@ permissions:
 jobs:
   build-and-publish:
     runs-on: ubuntu-latest
+    environment: protected-main-env
+    permissions:
+      id-token: write
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/pypi-release.yml
```diff
@@ -11,6 +11,9 @@ permissions:
 jobs:
   build-and-publish:
     runs-on: ubuntu-latest
+    environment: protected-main-env
+    permissions:
+      id-token: write
```
sobregosodd, line 16: why this?

ikretz: It's just part of what's needed for the setup.
See: https://docs.pypi.org/trusted-publishers/using-a-publisher/

PR #184, opened 2025-12-03
[SECRES-4132] Bugfix: npm installation target resolution

ikretz:
This PR fixes a bug in SCFW's npm installation target identification (`scfw run` command) in which targets resolved directly from a local `package-lock.json` file were not being included in the set of targets to verify.  It does this via a more nuanced understanding of which `--dry-run` verbose log lines are relevant to install target resolution.

To ensure complete isolation of the user's npm project from SCFW's internal interactions with `npm` for install target resolution, this PR creates a new class, `TemporaryNpmProject`, for SCFW to run its `npm` commands in.  The `package.json`, `package-lock.json` and `node_modules/` from the user's npm project (if any) are copied into the temporary one before commands are run.  This change was made because it was observed during testing that the `install --dry-run` option, in spite of its name, can nevertheless have side-effects on the user's project in certain situations that are visible to the user.

Finally, this PR overhauls the set of tests for npm, both for the `npm` CLI program itself as well as for the implementation of the `PackageManager` class for `npm`.  More precisely:

* Low-quality tests have been removed
* All prerequisite assumptions about `npm` CLI behavior made by the `Npm` class are checked
* The `Npm.resolve_install_targets()` method is tested in more diverse use-cases:
    - Empty directory
    - Fresh npm project (no installed dependencies)
    - npm projects with dependencies, with and without `package-lock.json` and with and without `node_modules/`

Other changes include:

* Refactor the Poetry tests to move common fixtures into a standalone module (to match what was done for npm)

### DIFF
--- a/tests/package_managers/npm_fixtures.py
+++ b/tests/package_managers/npm_fixtures.py
@@ -15,6 +15,9 @@ TEST_PACKAGE = "react"
 TEST_PACKAGE_LATEST = "18.3.0"
 TEST_PACKAGE_PREVIOUS = "18.2.0"
 
+TEST_PACKAGE_LATEST_SPEC = f"{TEST_PACKAGE}@{TEST_PACKAGE_LATEST}"
+TEST_PACKAGE_PREVIOUS_SPEC = f"{TEST_PACKAGE}@{TEST_PACKAGE_PREVIOUS}"
+
 
 @pytest.fixture
 def empty_directory():
@@ -31,7 +34,7 @@ def empty_directory():
 @pytest.fixture
 def new_npm_project():
     """
-    Initialize a new npm project in an empty directory.
+    Initialize a new npm project with no dependencies.
     """
     tempdir = TemporaryDirectory()
     tempdir_path = Path(tempdir.name)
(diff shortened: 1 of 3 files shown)


scfw/package_managers/npm/temp_project.py
```diff
@@ -0,0 +1,252 @@
...
+            if not (dependencies := lockfile.get("packages")):
+                raise KeyError("Missing dependencies data in package-lock.json")
+            if not (target_entry := dependencies.get(target_handle)):
+                raise KeyError(
+                    f"Missing entry for installation target {target_name} in package-lock.json"
+                )
+            if not (version := target_entry.get("version")):
+                raise KeyError(
+                    f"Missing version data for installation target {target_name} in package-lock.json"
+                )
```
tesnim5hamdouni, line 173: Are there still discrepancies between `node-modules` and `package-lock.json` after running the command?

tests/package_managers/test_pip.py
```diff
@@ -24,12 +22,34 @@ def pip_list() -> str:
     return subprocess.run(pip_list_command, check=True, text=True, capture_output=True).stdout.lower()
 
 
+def select_test_install_target(installed_packages: str) -> str:
+    """
+    Select a test target that is not in the given installed packages output.
+
+    This allows us to be certain when testing that nothing was installed in a dry-run.
```
tesnim5hamdouni, line 29: I can see why you were proud of the tests. Figuring out all these test cases is tricky 😄

tesnim5hamdouni (approved):
looks good to me, had one question
the temp_proj can be developed into a sandbox :D but all in due time

tests/package_managers/test_pip.py
ikretz: Yeah, speaking of which, in a future PR, I need to go back and overhaul similarly the PyPI tests.
Pip is a much simpler beast than npm but still, there's more we need to do here.

scfw/package_managers/npm/temp_project.py
ikretz: Good question, I wasn't able to detect any in testing.
It seems like if both are present then they are in agreement.

PR #191, opened 2025-12-15
[SECRES-4058] Add a default verifier for user-provided findings lists

ikretz:
This PR adds a new default verifier to SCFW for user-provided findings lists, expressed in YAML.

```bash
$ scfw run pip install pynamodb
Package pynamodb-6.1.0:
  - This package is licensed under an incompatible open source license and must not be used

The installation request was blocked. No changes have been made.

$ scfw run npm install axios
Package axios@1.13.2:
  - An internal fork of this package exists: please consider using it
```

Both `CRITICAL` and `WARNING` findings are supported, allowing for automatically blocking commands or merely warning users while giving them the option to continue.

Currently, the new verifier always looks for findings list YAML files in `$SCFW_HOME/list_verifier/`.  It is effectively disabled if this directory doesn't exist or exists but doesn't contain any YAML files.  Existing users of SCFW will not experience any differences in its behavior after upgrading without explicitly opting into the new verifier (i.e., by adding at least one YAML findings list in the appropriate directory).

A minimal example of the expected YAML format has been added to the repo at `examples/findings_list.yaml` directory.

Other changes include:
* Add tests for the new verifier
* Update the repo's third-party software documentation to include `PyYAML`, which is used for YAML parsing

### DIFF
--- a/examples/findings_list.yaml
+++ b/examples/findings_list.yaml
@@ -3,8 +3,10 @@ findings:
   - severity: CRITICAL
     finding: "This package is licensed under an incompatible open source license and must not be used"
     packages:
+      # Any version of PyPI package `foo` has this finding
       - ecosystem: PyPI
         name: foo
+      # Only the listed versions of npm package `bar` have this finding
       - ecosystem: npm
         name: bar
         versions:


scfw/verifiers/list_verifier/__init__.py
```diff
@@ -0,0 +1,94 @@
...
+            return (f for f in directory.glob(pattern) if f.is_file())
+
+        self._findings_map = FindingsMap()
+
+        if (scfw_home := os.getenv(SCFW_HOME_VAR)):
+            findings_lists_home = Path(scfw_home) / LIST_VERIFIER_HOME
+            if not findings_lists_home.is_dir():
+                return
+
+            findings_lists = itertools.chain(
```
sobregosodd, line 42: cool

scfw/verifiers/list_verifier/__init__.py
```diff
@@ -0,0 +1,94 @@
...
+
+    @classmethod
+    def supported_ecosystems(cls) -> set[ECOSYSTEM]:
+        """
+        Return the set of package ecosystems supported by `FindingsListVerifier`.
+
+        Returns:
+            The class' constant set of supported ecosystems: `{ECOSYSTEM.Npm, ECOSYSTEM.PyPI}`.
+        """
+        return {ECOSYSTEM.Npm, ECOSYSTEM.PyPI}
```
sobregosodd: maybe `for e in ECOSYSTEM` so we don't have to manually add new ones.
i don't see why we would want to skip new ones

scfw/verifiers/list_verifier/findings_map.py
```diff
@@ -0,0 +1,145 @@
...
+
+    def merge(self, other: Self):
+        """
+        Merge a second `FindingsMap` into the given one.
+
+        Args:
+            other:
+                The `FindingsMap` whose contents should be merged into `self`.
+
+                Note that no attempt is made to deduplicate the merged findings. That is,
```
tesnim5hamdouni, line 106: I think it's better to deduplicate to avoid unnecessary checks

sobregosodd (approved):

scfw/verifiers/list_verifier/findings_map.py
tesnim5hamdouni: Edit: on second thought, there is no easy way to deduplicate since the finding's "wording" can change and we can't trust the user to respect some self-declared finding id

tesnim5hamdouni (approved):
lgtm, left a question

scfw/verifiers/list_verifier/__init__.py
ikretz: True, I think for this verifier it makes sense to just auto-support any ecosystem.
For the others, since there's dedicated logic for each ecosystem, I think it makes sense add them one at a time.

scfw/verifiers/list_verifier/findings_map.py
ikretz: In the beginning, I was going to store sets of findings in the underlying map to handle deduplication automatically.
But as you essentially pointed out, it's unlikely that we end up in a situation where we have identical findings (same severity and message) coming from several lists.  So I just kept the lists.

scfw/verifiers/list_verifier/__init__.py
ikretz: Changed in [7d206af](https://github.com/DataDog/supply-chain-firewall/pull/191/commits/7d206af0edfc65781f89ebf49ede85370cf04a93)

PR #202, opened 2026-01-05
[SECRES-3526] Add a verifier for recently created packages

### LINKED ISSUE #180: Add optional warning to bypass if a package was released less than X hours ago
... could be a good setting to have, similarily to the [minimumReleaseAge](https://pnpm.io/settings#minimumreleaseage) setting of pnpm.

ikretz:
This PR adds a new default verifier to SCFW that warns on recently created packages:

```bash
$ SCFW_PACKAGE_MINIMUM_AGE=876000 scfw run npm install react
Package react@19.2.3:
  - Package react was created less than 876000 hours ago: treat new packages with caution
```

The environment variable `SCFW_PACKAGE_MINIMUM_AGE` allows the user to configure an integer number of hours expressing the minimum age they require packages to have, that is, the number of hours since package creation that they deem safe.  One day (24 hours) is used by default.  If a package was created (i.e., published to its package registry for the first time) less than this number of hours ago, a warning to this effect is raised.

Adding this verifier provides an additional level of protection to SCFW users against supply-chain attacks: all things being equal, recently-created packages are generally unvetted and should be treated with a greater deal of initial skepticism before being installed.

Other changes include:
* Update dev dependencies to the most recent major versions

Closes #180

### DIFF
--- a/scfw/verifiers/age_verifier/__init__.py
+++ b/scfw/verifiers/age_verifier/__init__.py
@@ -45,7 +45,6 @@ class PackageAgeVerifier(PackageVerifier):
                 user_minimum_age = int(m)
                 if user_minimum_age < 0:
                     raise ValueError("Minimum age cannot be negative")
-
                 minimum_age = user_minimum_age
             except Exception:
                 _log.warning(


scfw/verifiers/recency_verifier/__init__.py
```diff
@@ -0,0 +1,119 @@
...
+        """
+        try:
+            match package.ecosystem:
+                case ECOSYSTEM.Npm:
+                    creation_datetime_utc = npm.get_creation_datetime_utc(package.name)
+                case ECOSYSTEM.PyPI:
+                    creation_datetime_utc = pypi.get_creation_datetime_utc(package.name)
+
+            if datetime.now(tz=timezone.utc) - creation_datetime_utc < self.tolerance:
+                tolerance_hours = int(self.tolerance.total_seconds()) // 3600
```
sobregosodd: didn't know about `//` cool

scfw/verifiers/age_verifier/npm.py
```diff
@@ -0,0 +1,37 @@
...
+    Returns:
+        A UTC `datetime` representing the creation date of `package_name` on npm.
+
+    Raises:
+        requests.HTTPError: Failed to query the npm registry.
+        requests.exceptions.JSONDecodeError: Failed to parse query response as JSON.
+        RuntimeError: Package metadata missing required fields.
+        ValueError: Creation timestamp does not have the expected format.
+    """
+    r = requests.get(f"https://registry.npmjs.org/{package_name}")
```
sobregosodd, line 26: maybe we need to have some sort of cache for this

scfw/verifiers/age_verifier/pypi.py
```diff
@@ -0,0 +1,51 @@
...
+
+    Raises:
+        requests.HTTPError: Failed to query the PyPI registry.
+        requests.exceptions.JSONDecodeError: Failed to parse query response as JSON.
+        RuntimeError:
+            * Package metadata missing required fields.
+            * No upload timestamps found in package metadata.
+        ValueError: Creation timestamp does not have the expected format.
+    """
+    r = requests.get(f"https://pypi.org/pypi/{package_name}/json")
```
sobregosodd, line 28: same here

sobregosodd (approved):

README.md
```diff
@@ -17,6 +17,7 @@ By default, Supply-Chain Firewall includes verifiers for the following data sour
 
 - Datadog Security Research's public [malicious packages dataset](https://github.com/DataDog/malicious-software-packages-dataset)
 - [OSV.dev](https://osv.dev) advisories, both for malicious packages as well as vulnerabilities
+- Package registry metadata, warning when an installation target was created very recently
```
sobregosodd: Let's show the default here. 
Also don't we need to show the env variable somewhere in the readme?

ikretz: We have a whole bunch of configuration options for these verifiers now, too many to include in the README probably.  I have a generic note about "please refer to the API documentation" a couple lines down from here for full details on the options.

ikretz: Mentioned default tolerance to README in [b14dcd2](https://github.com/DataDog/supply-chain-firewall/pull/202/commits/b14dcd24fb57d4dd449d46ad3be6a7cb8c985dea)

scfw/verifiers/age_verifier/pypi.py
ikretz: See previous thread on caching.

scfw/verifiers/age_verifier/npm.py
ikretz: Caching is tough to get right for this because packages might be deleted and recreated.

When a package is deleted, its name goes back into circulation to be used again.  Each time a name is reused, we have to invalidate the cache and update the creation time.  But, since deletions and recreations can happen at any time, unless we check with the registry on every single request, we risk using an invalid creation date read from cache.  But then this defeats the purpose of caching.

All that said, package deletions are pretty rare events, but I'm not sure how to do caching in a way that accounts for deletions/recreations in a safe way.

sobregosodd: Let's think this through, maybe not in this PR. 
Options:
- We can use the E-Tag like we do for the manifest.json
- We can just invalidate cache if got it just a few days ago (e.g: 2 days), this would prevent the tool from rapidly requesting several times the same json, IMO is the more likely event. (if i try to install requests and it fails for whatever reason, i will retry in a few secs)
- We can store the cache in `package@version` format, so if `package@another_version` is requested cache will miss.

scfw/verifiers/age_verifier/__init__.py
```diff
@@ -0,0 +1,119 @@
...
+        too recent based on a user-configurable tolerance.
+
+        Args:
+            package: The `Package` to verify.
+
+        Returns:
+            A list containing a single `WARNING` finding if `package` is deemed to have
+            been created too recently, otherwise an empty list.
+        """
+        try:
```
sobregosodd, line 91: should we add a way / setting value here to disable this check?

scfw/verifiers/age_verifier/npm.py
ikretz: I hadn't thought of using the E-Tag for the metadata itself.  That seems like a promising approach.
Agree that we should revisit this in a subsequent PR.

scfw/verifiers/age_verifier/__init__.py
ikretz: I think we should opt-in users for this verifier by default.

The out-of-the-box settings (warn when a target is < 1 day old) are very conservative and, in that case, the warnings are of legitimate interest.

Down the road, I have plans to allow users to select and deselect verifiers on the command line by name, and at that time, they can opt out of this verifier, if desired.

Another, admittedly hacky, way to opt out right now is to set `SCFW_PACKAGE_RECENCY_TOLERANCE = 0`.

sobregosodd: Yes, but the `get_creation_datetime_utc` will execute and therefore the json request fetches

ikretz: Right, that's what we call per package verification.  I might not be following what you mean by "disable this check."

scfw/verifiers/age_verifier/__init__.py
```diff
@@ -0,0 +1,119 @@
...
+            match package.ecosystem:
+                case ECOSYSTEM.Npm:
+                    creation_datetime_utc = npm.get_creation_datetime_utc(package.name)
+                case ECOSYSTEM.PyPI:
+                    creation_datetime_utc = pypi.get_creation_datetime_utc(package.name)
```
sobregosodd, line 96: I PREFER here instead,
```suggestion
            match package.ecosystem:
                case ECOSYSTEM.Npm:
                    verifier = npm
                case ECOSYSTEM.PyPI:
                    verifier = pypi
            creation_datetime_utc = verifier.get_creation_datetime_utc(package.name)
```

scfw/verifiers/age_verifier/__init__.py
sobregosodd: Sorry, I mean, IMO we should add a top most condition saying:
```
if self.tolerance == 0: return []
```
So we don't waste time doing pointless `get_creation_datetime_utc`

ikretz: Ahh!  Gotcha, great idea.  Added in [4fd606c](https://github.com/DataDog/supply-chain-firewall/pull/202/commits/4fd606c796afbe9af0cc8e62ea8e93b316c645c9).

scfw/verifiers/age_verifier/__init__.py
ikretz: But then the the functions are required to have the same name.  Which we could take care to ensure, but it's an extra constraint that we don't really need.

ikretz: We could do:

```python
match package.ecosystem:
    case ECOSYSTEM.Npm:
        get_creation_datetime_utc = npm.get_creation_datetime_utc
    case ECOSYSTEM.PyPI:
        get_creation_datetime_utc = pypi.get_creation_datetime_utc

creation_datetime_utc = get_creation_datetime_utc(package.name)
```
?

sobregosodd: is just a preference, go with whatever your heart says

PR #258, opened 2026-04-15
[SECRES-4731] Minimum age verifier looks at release instead of creation dates

ikretz:
This PR enhances the package minimum age verifier to include the specific version number in its analyses.  Now, the particular release publication date is analyzed instead of the package absolute creation date.  This is being done to enhance the utility of this verifier against instances of package takeover such as the recent [axios](https://github.com/axios/axios/issues/10636) compromise.

Other changes include:

* Add matrix tests for the latest npm releases


scfw/verifiers/age_verifier/pypi.py
```diff
@@ -8,42 +8,39 @@
...
-            upload_timestamp = metadata.get("upload_time_iso_8601")
-            if not upload_timestamp:
-                continue
+    release_datetimes = set()
+    for metadata in release_metadata:
+        if (release_datetime := metadata.get("upload_time_iso_8601")):
+            release_datetimes.add(datetime_parser.parse(release_datetime))
 
-            upload_datetimes.add(datetime_parser.parse(upload_timestamp))
+    if not release_datetimes:
```
tesnim5hamdouni, line 43: Are there cases where release dates aren't present? is there a better way to gracefully handle their absence (only raise a warning but continue the execution?)

tesnim5hamdouni (approved):
lgtm, left a question

scfw/verifiers/age_verifier/pypi.py
ikretz: It _should_ not ever happen, but if it did, we raise an error here and it gets converted into a logged warning in `verifiers/age_verifier/__init__.py`, so it's non-fatal.

PR #267, opened 2026-05-28
[SECRES-4877] Use package source attribute in npm package manager

ikretz:
This PR wires up `Package.source` (added in #259) for the `Npm` package manager so every npm `Package` carries source provenance (remote URL or local path).

In adding this feature, it was discovered that simply copying the `package.json`, `package-lock.json` and `node_modules/` from the original project directory into the `TemporaryNpmProject` directory is insufficient.  Specifically, npm prefers to reference local package dependencies via relative (to the project directory) file paths. Thus, simply copying these files into a new directory breaks these references.  The context manager entry code has been significantly enhanced so that these previously dangling references are re-relativized with respect to the temporary directory.

Other changes include:

  - A one-line `if` → `elif` fix in `pip.py` where remote pip packages were having their `RemotePackageSource` immediately overwritten with `None`.
  - Additional tests for the new behaviors
  - Update dependencies to their latest compatible versions


scfw/verifiers/osv_verifier/__init__.py
```diff
@@ -135,7 +135,7 @@ def failure_message() -> str:
             return [(FindingSeverity.WARNING, f"Package ecosystem {package.ecosystem} is not supported")]
 
         vulns = []
-        query = {
+        query: dict[str, str | dict[str, str]] = {
```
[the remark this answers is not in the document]
ikretz: Added in [64bbb6f](https://github.com/DataDog/supply-chain-firewall/pull/267/commits/64bbb6f6e86140418a4fbf38b0d392997df6d471)

scfw/package_managers/npm/temp_project.py
```diff
@@ -121,47 +239,46 @@ def resolve_install_command_targets(self, install_command: list[str]) -> list[Pa
...
+
+            # Some versions of npm report CHANGE for local dependencies already present in
+            # `node_modules` when the resolved path is re-relativized for the temp project
+            # Since they are not being freshly downloaded or installed, they may be excluded
+            return [
+                handle for handle in target_handles
+                if not (
+                    pre_install_packages.get(handle, {}).get("resolved", "").startswith(_LOCAL_DEPENDENCY_PREFIX)
+                    and (temp_dir_path / handle).exists()
+                )
```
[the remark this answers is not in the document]
ikretz: Changed in [64bbb6f](https://github.com/DataDog/supply-chain-firewall/pull/267/commits/64bbb6f6e86140418a4fbf38b0d392997df6d471)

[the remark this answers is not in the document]
ikretz: The suggested change is necessary but so is the original check: they are complimentary.
Changed in [b263d0a](https://github.com/DataDog/supply-chain-firewall/pull/267/commits/b263d0ae2a08f163c08d54c8b257fc775cad7b85)

scfw/package_managers/npm/temp_project.py
```diff
@@ -231,14 +381,8 @@ def handle_to_package(
             if not (dependencies := json.load(f).get("packages")):
```
tesnim5hamdouni: Claude: An empty packages dict {} is falsy in Python, so this will raise a KeyError even when the lockfile is valid
  but has no packages. This breaks the case where a project has a valid lockfile with zero dependencies.
  
  Fix: Check for None explicitly:
```
  if (dependencies := json.load(f).get("packages")) is None:
      raise KeyError("Missing dependencies data in package-lock.json")
```

scfw/package_managers/npm/temp_project.py
```diff
@@ -121,47 +239,54 @@ def resolve_install_command_targets(self, install_command: list[str]) -> list[Pa
...
+                    and line_tokens[2] in {"ADD", "CHANGE"}
+                ):
                     target_handles.append(line_tokens[3])
 
-            return target_handles
+            lockfile_path = temp_dir_path / "package-lock.json"
+            if not lockfile_path.is_file():
+                return target_handles
+            with open(lockfile_path) as f:
+                pre_install_packages = json.load(f).get("packages", {})
```
tesnim5hamdouni, line 264: Claude Nit:
  The lockfile is loaded twice:
  - Line 264: To extract pre-install packages for filtering
  - Line 381: To get dependencies for resolving targets
  
  This is inefficient for large lockfiles. The first load is only used to check for link: true entries. Consider
  restructuring to reuse the parse result.
  
  Suggestion: Load once and cache the result, or defer the first check until after the actual install.

ikretz: This is true, but I think there's a good reason to do it: we don't want to run the `npm install` command unless we know we really have to.  In order to get this down to one read/parse of `package-lock.json`, we would have to unconditionally run the command and then do this filtering step afterwards.  Moreover, from an efficiency standpoint, it's not clear (at least not to me) that replacing an unnecessary JSON parse with an unnecessary `npm` execution saves us anything.

What do you think?

scfw/package_managers/npm/temp_project.py
ikretz: Nice catch.

sobregosodd (approved):

scfw/package_managers/npm/temp_project.py
ikretz: Fixed in [165352e](https://github.com/DataDog/supply-chain-firewall/pull/267/commits/165352e808dec499c16aae85ae5dc738402e559f)

scfw/package_managers/npm/temp_project.py
tesnim5hamdouni: I think that makes sense

tesnim5hamdouni (approved):
:shipit:

PR #286, opened 2026-07-01
Tham/poetry dep confusion

tesnim5hamdouni:
Add dependency confusion support to poetry.
Approach:
scan for poetry.lock file in the project. If present, we classify the `[package.source]` entries based on their type. `file` and `directory` types are `LocalPackageSource`, everything else (`git`, `ssh`, `legace`, `[[package]]` -ie pypi) is considered remote.
If absent, we copy the `pyproject.toml` file to a temp folder and use the `lock` command to generate the lock file. No considerations for virtual envs, or versions are needed!


pyproject.toml
```diff
@@ -9,6 +9,7 @@ dependencies = [
     "python-dotenv",
     "pyyaml",
     "requests",
+    "tomli>=2.0; python_version<'3.11'",
```
ikretz, line 12: Can you also update the `LICENSE-3rdparty.csv` file with the new dependency?

tesnim5hamdouni: @datadog fix this

scfw/package_managers/poetry.py
```diff
@@ -201,3 +220,144 @@ def _normalize_command(self, command: list[str]) -> list[str]:
...
+
+        with TemporaryPoetryProject(executable, project_dir) as generated_lock:
+            return _parse_lock_file(generated_lock) if generated_lock else {}
+
+    except Exception as e:
+        _log.warning("Could not determine package sources from poetry.lock: %s", e, exc_info=True)
+        return {}
+
+
+class TemporaryPoetryProject:
```
ikretz: nit: could you break this out into a separate module, e.g., `scfw/package_managers/poetry/temp_project.py`?

You can then rename `scfw/package_managers/poetry.py` --> `scfw/package_managers/poetry/__init__.py`.

Be sure also to add `"scfw.package_managers.poetry"` into `pyproject.toml` under `packages` (take it from someone who has forgotten in the past)

scfw/package_managers/poetry.py
```diff
@@ -201,3 +220,144 @@ def _normalize_command(self, command: list[str]) -> list[str]:
...
+
+        Args:
+            executable: Path to the `poetry` executable.
+            project_dir: Path to the project directory containing `pyproject.toml`.
+        """
+        self._executable = executable
+        self._project_dir = project_dir
+        self._temp_dir: Optional[TemporaryDirectory] = None
+
+    def __enter__(self) -> Optional[Path]:
```
ikretz: Take a look at `TemporaryNpmProject` in `./npm/temp_project.py`.
You want a separate, identical clone of the given project that you can run whatever commands you want in.

scfw/package_managers/poetry.py
```diff
@@ -201,3 +220,144 @@ def _normalize_command(self, command: list[str]) -> list[str]:
...
+        temp_path = Path(self._temp_dir.name)
+
+        shutil.copy(self._project_dir / "pyproject.toml", temp_path / "pyproject.toml")
+        if (poetry_toml := self._project_dir / "poetry.toml").is_file():
+            shutil.copy(poetry_toml, temp_path / "poetry.toml")
+
+        _log.warning("No poetry.lock found; generating one in a temporary directory (this may be slow)")
+
+        result = subprocess.run(
+            [self._executable, "lock"],
```
ikretz: From Claude:

`poetry add` new packages always get `source=None` — The most security-relevant finding: `TemporaryPoetryProject` copies the pre-`add` `pyproject.toml`, so the generated lock file won't contain the new package either. Both code paths miss newly-added packages. The test on line 76 explicitly acknowledges this with a comment rather than an assertion.

The core problem is a timing mismatch: the source map is built from the current lock file, but `poetry add` is installing a package that isn't in the lock file yet.

What happens step by step with `poetry add requests`:

1. The firewall intercepts the command before Poetry runs.
2. `resolve_install_targets` runs `poetry add requests --dry-run`, which reports that `requests==2.31.0` will be installed.
3. It then calls `_get_source_map` to figure out where requests comes from (PyPI? a private index?).
4. `_get_source_map` looks for `poetry.lock` in the project directory. It exists (the project has other deps), so it reads it directly.
5. `requests` is not in `poetry.lock` — it's a new dependency. So `source_map.get(("requests", "2.31.0"))` returns `None`.
6. `resolve_install_targets` returns `Package(ecosystem=PyPI, name="requests", version="2.31.0", source=None)`.

Why the `TemporaryPoetryProject` fallback doesn't help:

The fallback path (when no lock file exists) runs `poetry lock` on a **copy** of `pyproject.toml`. But this copy is made before `poetry add` has modified `pyproject.toml` to declare `requests` as a dependency. So the generated lock file also won't contain `requests`. The fallback was designed for projects with no lock file at all — it doesn't solve the `add` case.

Why this matters:

`source=None` is the firewall's signal that it couldn't determine provenance. All verifiers treat it as "assume registry, proceed with warning." In a dependency confusion attack, an attacker publishes a malicious package to PyPI under the same name as an internal package. When a developer runs `poetry add internal-tool`, the firewall should detect that `internal-tool` is expected to come from a private index — but it gets `source=None` and waves it through.

ikretz: One way to fix this (I think) would be to change `TemporaryPoetryProject` to not do anything with `poetry lock` in the `__enter__` method and then running `poetry add requests --lockfile-only` (or whatever the right option is) in the temp environment, then you read out the source map from that.

scfw/package_managers/poetry.py
```diff
@@ -201,3 +220,144 @@ def _normalize_command(self, command: list[str]) -> list[str]:
...
+        """
+        Set up the temporary project and generate a `poetry.lock` file.
+
+        Returns:
+            The `Path` to the generated `poetry.lock`, or `None` if generation failed.
+        """
+        self._temp_dir = TemporaryDirectory()
+        temp_path = Path(self._temp_dir.name)
+
+        shutil.copy(self._project_dir / "pyproject.toml", temp_path / "pyproject.toml")
```
ikretz: Per Claude (I'm summarizing): If you error out at this step, you leak `self._temp_dir`.
Wrap in a `try-except` block and manually free `self._temp_dir` if anything goes wrong.

scfw/package_managers/poetry.py
```diff
@@ -201,3 +220,144 @@ def _normalize_command(self, command: list[str]) -> list[str]:
...
+        _log.warning("Could not determine package sources from poetry.lock: %s", e, exc_info=True)
+        return {}
+
+
+class TemporaryPoetryProject:
+    """
+    Prepares a temporary Poetry project that duplicates a given one, allowing a
+    `poetry.lock` file to be generated without modifying the original project.
+    """
+    def __init__(self, executable: str, project_dir: Path):
```
ikretz: From Claude: Do more logging in this class 😁

ikretz (changes requested):
looks good overall, passing along some stuff from Claude and that I noticed.

scfw/package_managers/poetry.py
tesnim5hamdouni: from what i understand, most of npm complexity (copying the installed state, node modules, etc.) comes from the absence of a `--dry-run`

tesnim5hamdouni: changed `__enter__` to return path to the temp directory instead of the lock file

scfw/package_managers/poetry/temp_project.py
```diff
@@ -0,0 +1,219 @@
...
+        lock_data = tomllib.load(f)
+
+    source_map: dict[tuple[str, str], LocalPackageSource | RemotePackageSource] = {}
+    for package in lock_data.get("package", []):
+        name = package.get("name", "")
+        version = package.get("version", "")
+        if not name or not version:
+            continue
+
+        if source := package.get("source"):
```
ikretz, line 234: Via Claude:

## Primary/private-index packages get silently trusted as PyPI

The `_parse_lock_file` function reads `poetry.lock` and, for each locked package, decides where it "came from" so that later code (the verifiers in `scfw/verifiers/*`) can check `Package.has_registry_source()` — is this package genuinely from the real PyPI registry (pypi.org / files.pythonhosted.org), or from somewhere else that can't be trusted the same way?

The logic hinges entirely on whether the package's `[[package]]` entry in `poetry.lock` has a `[package.source]` sub-table:

- Has a `[package.source]` table → use its url to build a `RemotePackageSource` (or `LocalPackageSource` for path/file deps).
- No `[package.source]` table at all → the else branch assumes this means "must be PyPI" and fabricates `https://pypi.org/project/{name}/{version}/`.

### Why that assumption is wrong

Poetry only writes `[package.source]` when a package's source differs from the project's implicit default resolver. If a project configures a custom index as the primary/default source — e.g.:

```toml
[[tool.poetry.source]]
name = "internal"
url = "https://artifactory.internal.company.com/api/pypi/simple/"
priority = "primary"
```

(or the older, now-deprecated `default = true` form) — then that internal index becomes the implicit default, the same role PyPI normally plays. Every package resolved from it, including ordinary third-party packages like `requests` or `numpy`, gets locked without a `[package.source]` table, because as far as Poetry's lock-writer is concerned, it came from "the default source" — no annotation needed.

So the else branch in `_parse_lock_file` cannot actually distinguish "no source table because it's really PyPI" from "no source table because it's the project's private primary index." It assumes the former unconditionally.

The fix is to stop assuming and instead ask: what is this project's actual default source?

### Fix

1. Determine the project's configured primary/default source, by reading `pyproject.toml` (already sitting next to the lock file we're parsing — real project or temp copy) for `[[tool.poetry.source]]` entries with `priority = "primary"` (current syntax) or the legacy `default = true`. Poetry allows at most one such entry per project.

```python
def _get_primary_source_url(project_dir: Path) -> Optional[str]:
    pyproject_path = project_dir / "pyproject.toml"
    if not pyproject_path.is_file():
        return None
    with open(pyproject_path, "rb") as f:
        data = tomllib.load(f)
    sources = data.get("tool", {}).get("poetry", {}).get("source", [])
    for src in sources:
        if src.get("priority") == "primary" or src.get("default") is True:
            return src.get("url")
    return None
```

2. Thread that into `_parse_lock_file` so the "no source table" branch stops hardcoding PyPI and instead falls back to the primary source when one exists:

```python
def _parse_lock_file(lock_path: Path, primary_source_url: Optional[str] = None) -> dict[...]:
    ...
    else:
        if primary_source_url:
            source_map[(name, version)] = RemotePackageSource(primary_source_url)
        else:
            source_map[(name, version)] = RemotePackageSource(
                f"{_PYPI_PROJECT_BASE_URL}/{name}/{version}/"
            )
```

3. Update the two call sites (`get_source_map`, `resolve_via_lock`) to compute `primary_source_url = _get_primary_source_url(project_dir)` once and pass it through to every `_parse_lock_file` call. Both already have project_dir/temp_dir in scope, so this is a small threading change, not a new mechanism.

### Why this is the right fix, not a workaround

Once a package resolves to the primary source's URL instead of pypi.org, `has_registry_source()` correctly returns `False` for it (the URL doesn't match `pypi.org/files.pythonhosted.org`), which causes the verifiers to raise `UnverifiablePackage` — the intended fail-safe behavior for a package whose provenance the tool can't independently confirm. That's consistent with how `pip.py` already treats packages from custom/private indexes, so this brings poetry's behavior in line with the rest of the codebase rather than inventing new semantics.

If the project has no configured primary source, the original PyPI-fallback behavior is preserved exactly as before — this only changes behavior for the specific case that's actually ambiguous today.

scfw/package_managers/poetry/temp_project.py
```diff
@@ -0,0 +1,219 @@
...
+        """
+        Initialize a new `TemporaryPoetryProject`.
+
+        Args:
+            project_dir: Path to the project directory containing `pyproject.toml`.
+        """
+        self._project_dir = project_dir
+        self._temp_dir: Optional[TemporaryDirectory] = None
+
+    def __enter__(self) -> Path:
```
ikretz, line 163: Via Claude:

## Local path dependencies make add/update bypass verification entirely

A very common Poetry project layout declares a dependency on a sibling package by filesystem path:

```toml
[tool.poetry.dependencies]
sibling = {path = "../sibling-pkg", develop = true}
```

This is standard for monorepos. When `poetry add/update --lock` runs in the temp copy, Poetry tries to resolve sibling by looking at `../sibling-pkg` relative to the temp directory — but that directory was never copied in. Poetry fails immediately with something like `Path ... for sibling does not exist`, exit code 1.

### Fix

The fix is simpler than trying to enumerate and copy every possible local reference: place the temp directory as a sibling of the real project directory instead of in the system temp root.

```python
self._temp_dir = TemporaryDirectory(prefix=".scfw-poetry-", dir=self._project_dir.parent)
```

Why this works for any depth of relative path, not just the one-level case: Poetry resolves `path = "../sibling-pkg"` relative to the directory containing `pyproject.toml`. If the temp directory has the same parent as the real project directory, then for any relative path `../../../foo`, the first `..` from both directories lands on the identical shared parent — and every subsequent `../descent` from that shared point is identical regardless of which sibling you started in. So sibling placement makes relative path dependencies resolve to the real, correct location automatically, with no need to enumerate or copy the referenced directories at all. Absolute path dependencies already work regardless of temp-dir location, so this doesn't regress them.

This only fixes references that are relative to a parent directory (path/file: dependencies). It would not fix a same-directory reference (e.g., a plugin reading `.git/` or include globs from the project root itself) — but the earlier verification already ruled out readme as a trigger, and path dependencies are the confirmed, common case this closes.

ikretz: You could also rewrite the relative paths in the copied files the way `Npm.__enter__` does.
This solution is much slicker, feel free to choose either.

scfw/package_managers/poetry/temp_project.py
```diff
@@ -0,0 +1,219 @@
...
+    with TemporaryPoetryProject(project_dir) as temp_dir:
+        lock_path = temp_dir / "poetry.lock"
+        old_map = _parse_lock_file(lock_path) if lock_path.is_file() else {}
+
+        lock_command = _strip_directory_flag(command) + ["--lock"]
+        subprocess.run(lock_command, cwd=temp_dir, check=True, text=True, capture_output=True)
+
+        new_map = _parse_lock_file(lock_path) if lock_path.is_file() else {}
+        return {key: source for key, source in new_map.items() if key not in old_map}
+
```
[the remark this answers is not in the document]
ikretz: Claude also found this one:

## poetry update misses drift between the lock file and the installed environment

Before this PR, `add/update` went through the same `--dry-run` mechanism as `install/sync`: run the real command with `--dry-run` and parse Poetry's own "Installing/Updating/Downgrading" lines from stdout. Poetry's `--dry-run` output for `update` reflects a comparison between what's currently installed in the environment and what the command would newly resolve to.

The new code replaces that with `resolve_via_lock`, which instead:
1. Reads the project's existing `poetry.lock` → `old_map` (a `{(name, version): source}` mapping).
2. Runs the real `add/update --lock` command in a temp copy, which regenerates `poetry.lock` there.
3. Reads the regenerated lock → `new_map`.
4. Returns whatever keys exist in `new_map` but not in `old_map`.

This is a lock-vs-lock diff, comparing the file to itself before and after re-locking. It never reads the installed environment (`poetry show`, `site-packages`, etc.) at any point.

`(name, version)` is the diff key. If a package's version is unchanged between `old_map` and `new_map`, that key is excluded from the result — the code concludes "nothing new here."

But "the lock file didn't change" and "nothing needs to be installed" are not the same fact. Consider: a project's `poetry.lock` already pins `requests==2.31.0` — maybe it was regenerated separately, checked in by someone else, or the venv was never synced after a prior `poetry lock` — but the actual installed environment still has `requests==2.28.0`. Running the real `poetry update` in this state:
- Resolves dependencies again, finds `requests==2.31.0` is already what the lock says → the lock file itself doesn't change.
- But Poetry's installer step (which the new code doesn't even reach — it stops at `--lock`) would upgrade the installed requests from 2.28.0 to 2.31.0 to bring the environment in sync with the lock.

`resolve_via_lock`'s diff sees `old_map[("requests", "2.31.0")]` already present (since the lock didn't change) and therefore reports nothing as an install target for that package — even though the real command changes what's actually installed.

How this was confirmed, not just theorized

This exact scenario was previously tested. Before the PR, test_poetry_command_resolve_install_targets_update ran three fixture-driven cases, including poetry_project_target_previous_lock_latest — a project fixture explicitly constructed with the environment at the previous version but the lock already at the latest version. Against that fixture, the old --dry-run-based code correctly reported the latest version as an install target for update.

The new test (test_poetry_command_resolve_install_targets_update) only exercises one scenario — poetry_project_target_previous_loose_constraint, where the lock file itself changes on update (a case the new diff-based approach does handle correctly). The old fixture is still defined in poetry_fixtures.py but is no longer passed to any update test, so the env/lock-divergence scenario it was built to catch is silently untested — and, per the trace above, the new implementation actually fails it.

### Fix 

The core defect is that resolve_via_lock's baseline for "already satisfied" is only "was this (name, version) in the old lock file" — it should be "was this (name, version) in the old lock file and actually installed in the environment." Those two conditions coincide in the common case (env stays in sync with the lock), which is why the new single test case passes, but they diverge exactly in the scenario Finding 3 describes.

The fix

Compute the installed-packages set once (Poetry already has a method for this — `get_installed_packages()`, backed by `poetry show --all`) and use it to filter what counts as "already satisfied" before diffing:

```python
# scfw/package_managers/poetry/temp_project.py
def resolve_via_lock(
    command: list[str], installed: set[tuple[str, str]]
) -> dict[tuple[str, str], LocalPackageSource | RemotePackageSource]:
    project_dir = _resolve_project_dir(command)

    with TemporaryPoetryProject(project_dir) as temp_dir:
        lock_path = temp_dir / "poetry.lock"
        old_map = _parse_lock_file(lock_path) if lock_path.is_file() else {}

        lock_command = _strip_directory_flag(command) + ["--lock"]
        subprocess.run(lock_command, cwd=temp_dir, check=True, text=True, capture_output=True)

        new_map = _parse_lock_file(lock_path) if lock_path.is_file() else {}
        # Only packages already correctly reflected in BOTH the old lock and the
        # installed environment count as "no change needed" — a package pinned
        # in the lock but not yet installed at that version must still be reported.
        satisfied = {key for key in old_map if key in installed}
        return {key: source for key, source in new_map.items() if key not in satisfied}
```

```python
# scfw/package_managers/poetry/__init__.py
if any(subcommand in command for subcommand in LOCK_REGEN_SUBCOMMANDS):
    installed = {(p.name, p.version) for p in self.get_installed_packages()}
    source_map = resolve_via_lock(command, installed)
    ...
```

#### Why this restores correctness without regressing the fix this PR made

Walk through both scenarios with `satisfied = {key in old_map if key in installed}`:

- Env already matches the lock (the common case, and the only case the new test covers): every key in `old_map` is also in `installed`, so `satisfied == old_map`, and the result is identical to the current `key not in old_map` behavior — no regression.
- Lock is ahead of the environment (Finding 3's scenario — `requests==2.31.0` in the lock, 2.28.0 installed): `("requests", "2.31.0")` is in `old_map` but not in `installed` (env has a different version), so it's excluded from `satisfied`. Since `new_map` still contains `("requests", "2.31.0")` (the re-lock resolves to the same pin), it survives the filter and is correctly reported as a target — matching what the old `--dry-run`-based code did, and what the deleted `poetry_project_target_previous_lock_latest` fixture asserted.
- A genuinely new/changed dependency: same as today — the key doesn't exist in `old_map` at all, so it's never in `satisfied`, and it's reported regardless.

#### Cost and test coverage

`get_installed_packages()` is one `poetry show --all` subprocess call — cheap relative to the `--lock` resolution that already dominates this code path, so it doesn't meaningfully change the latency profile discussed earlier.

To close the regression for good, I'd also restore the deleted `poetry_project_target_previous_lock_latest` fixture as a case in `test_poetry_command_resolve_install_targets_update`, so this exact scenario is exercised in CI again rather than relying on manual review to catch a reintroduction.

scfw/package_managers/poetry/temp_project.py
```diff
@@ -0,0 +1,219 @@
...
+
+def _resolve_project_dir(command: list[str]) -> Path:
```
ikretz, line 126: Via Claude:

## `--directory=value` silently resolves the wrong project directory

Both functions look for `"--directory"` as an exact, standalone token in command, immediately followed by a separate token holding the path — i.e. they only handle `poetry --directory /some/path add pkg`. They don't handle the equally valid, attached form: `poetry --directory=/some/path add pkg`.

When the attached form is used:
- `command.index("--directory")` raises `ValueError` (no such literal token exists — the token is `"--directory=/some/path"`), so `_resolve_project_dir` falls through to `Path.cwd()`.
- `_strip_directory_flag`'s `if flag in stripped` check also misses, so the flag+value stays in the command unmodified when it's later re-run against the temp directory.

### Fix

The bug exists because two different functions (`_resolve_project_dir` and `_strip_directory_flag`) each independently scan command for a literal `"--directory"/"-C"` token, and both share the same blind spot: neither handles the `--directory=value` attached form. Patching each one to add another if branch would just duplicate the fix in two places — the same drift risk the reuse pass on this diff also called out (these two functions already had slightly different scanning logic for the same flag).

The better fix is to normalize the command once, before either function ever sees it, so their existing "look for a standalone token" logic doesn't need to change at all.

#### Where to put the normalization

`Poetry._normalize_command` (`scfw/package_managers/poetry/__init__.py`) is already the single place every command passes through before any temp-project logic runs (it's called at the top of `resolve_install_targets`). Extending it there fixes every downstream consumer — `_resolve_project_dir`, `_strip_directory_flag`, and even the early `--dry-run/-h/--version` checks — with one change:

```python
def _normalize_command(self, command: list[str]) -> list[str]:
    if not command:
        raise ValueError("Received empty poetry command line")
    if command[0] != self.name():
        raise ValueError("Received invalid poetry command line")

    normalized = [self._executable]
    for token in command[1:]:
        if token.startswith("--directory=") or token.startswith("-C="):
            flag, _, value = token.partition("=")
            normalized.extend([flag, value])
        else:
            normalized.append(token)
    return normalized
```

This splits `--directory=/some/path` into `["--directory", "/some/path"]` before anything else runs. `_resolve_project_dir` and `_strip_directory_flag` then find the flag exactly the way they already expect to, with zero changes to either function.

scfw/package_managers/poetry/temp_project.py
```diff
@@ -0,0 +1,219 @@
...
+    for package in lock_data.get("package", []):
+        name = package.get("name", "")
+        version = package.get("version", "")
+        if not name or not version:
+            continue
+
+        if source := package.get("source"):
+            url = source.get("url")
+            if source.get("type") in {"directory", "file"} and url:
+                source_map[(name, version)] = LocalPackageSource(Path(url))
```
ikretz: From Claude:

Resolve the `url` to an absolute path to avoid confusion about what relative paths are relative to:

```python
if source.get("type") in {"directory", "file"} and url:
    source_map[(name, version)] = LocalPackageSource((lock_path.parent / url).resolve())
```

scfw/package_managers/poetry/temp_project.py
tesnim5hamdouni: no op

scfw/package_managers/poetry/temp_project.py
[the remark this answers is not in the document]
tesnim5hamdouni: actually poetry show still shows what's in the lock file instead of the installation environment.
test: 
`poetry add pynamodb`
`poetry show pynamodb` gives 6.1.0, same thing as `pip show pynamodb`
then `pip install -Iv pynamodb==6.0.2` + `pip show pynamodb` shows 6.0.2, 
but `poetry show` still gives 6.1.0!

scfw/package_managers/poetry/temp_project.py
tesnim5hamdouni: it turns out the bug turns deeper than this. the `resolve_project_dir` falls back to the `cwd` even if `--directory` is set because we're not matching the `--directory=/path`
thus we end up copying the wrong directory into the temp folder 🤦‍♀️

ikretz (approved):
:shipit:

PR #293, opened 2026-07-17
Report `install_timestamp` in `/scfw/report` payload

dastrong:
## Summary
- Capture a timestamp as the first statement of `run_firewall()` (`scfw/firewall.py`), before package-manager resolution, verification, or any interactive prompt — the closest available proxy to when the user actually triggered the install, rather than when the report is later assembled.
- Thread it through as a new `timestamp: datetime` field on `FirewallRunSummary` (`scfw/logger.py`).
- `DDCodeSecurityLogger.generate_api_payload()` (`scfw/loggers/dd_codesec_logger.py`) now reports it as `attributes.install_timestamp` (ISO 8601, UTC) in the payload sent to `/scfw/report`, matching the schema defined in K9CODESEC-3653.
- Added `tests/test_dd_codesec_logger.py`, which had no prior coverage: verifies `install_timestamp` passthrough and the ALLOW/BLOCK report-shaping logic.

Ref: K9CODESEC-3702

## Test plan
- [x] Added unit tests asserting `generate_api_payload()` output includes `install_timestamp` as an ISO 8601 UTC string
- [x] `pytest tests/test_dd_codesec_logger.py tests/test_firewall.py tests/test_report.py` passes (57 passed)
- [x] Confirmed by construction that the timestamp is captured before verification/prompt, not at payload-build time


tests/test_dd_codesec_logger.py
```diff
@@ -0,0 +1,79 @@
+"""
+Tests of the Datadog Code Security API logger.
```
ikretz, line 2: Nice! Could you wire this into the Makefile (maybe a `test-loggers` recipe) and the CI `test` workflow?

dastrong: Good call — added a `test-loggers` Makefile recipe and matching CI job, following the existing per-file pattern. Pushed in ab22177.

ikretz (approved):
Looks great!

PR #300, opened 2026-08-20
:sparkles: Add scfw doctor command to troubleshoot installations

marcwieserdev:
## 🚀 Motivation
Make SCFW installation issues easier to diagnose by checking credentials and package-manager aliases from one command.

## 📚 Documentation
**Document** | **Link or Detail**
-------------|------------------
RFC | N/A
Incident | N/A
Jira Ticket | [Add a scfw doctor command to troubleshoot install issues](https://datadoghq.atlassian.net/browse/K9CODESEC-5028)

## 📝 Summary
Adds a `scfw doctor` command that reports API and application key sources, inspects SCFW-managed shell aliases, and explains when the current terminal may need to be reloaded. Refactors credential resolution to expose its source, adds managed-block parsing and focused coverage, and documents repository package responsibilities and release-build validation in `AGENTS.md`.

## 🧪 Testing
- [x] New tests were added for new logic.
- [x] Existing tests were updated for new logic, and not only so that they pass!
- [ ] Benchmark results prove that performance is the same or better.

## 🚧 Staging validation
- [x] Proof that it works as expected, including profiling or UX screenshots.

image

## 🆘 Recovery
Notes for on-call - **select only one**:
- [x] The change can be rolled back.
- [ ] Do not roll back. Why?:

### DIFF
--- a/scfw/internal/cli/doctor.go
+++ b/scfw/internal/cli/doctor.go
@@ -5,7 +5,12 @@
 
 package cli
 
-import "github.com/spf13/cobra"
+import (
+	"fmt"
+
+	"github.com/DataDog/supply-chain-firewall/scfw/internal/ddapi"
+	"github.com/spf13/cobra"
+)
 
 var doctorCmd = &cobra.Command{
 	Use:   "doctor",
@@ -16,5 +21,21 @@ var doctorCmd = &cobra.Command{
 
 func runDoctor(cmd *cobra.Command, args []string) error {
 	// Installation and configuration checks will be added here.
-	return nil
+	apiKey, appKey, apiKeyCredentialSource, appKeyCredentialSource, err := ddapi.DDCredentials()
+
+	if err != nil {
+		fmt.Printf("❌ Credentials not found with an error : %v\n", err)
+	} else {
+		if len(apiKey) == 0 {
+			fmt.Printf("❌ API Key not found in keychain nor environment\n")
+		} else {
+			fmt.Printf("✅ API Key found in %s\n", apiKeyCredentialSource)
+		}
+		if len(appKey) == 0 {
+			fmt.Printf("❌ Application Key not found in keychain nor environment\n")
+		} else {
+			fmt.Printf("✅ Application Key found in %s\n", appKeyCredentialSource)
+		}
+	}
+	return err
 }
--- a/scfw/internal/cli/doctor.go
+++ b/scfw/internal/cli/doctor.go
@@ -75,12 +75,18 @@ func reportAliases(cmd *cobra.Command, home string) error {
 	aliasLocations := make(map[string][]string)
 	invalidAliases := make(map[string][]invalidAlias)
 	var errs []error
+	out := cmd.OutOrStdout()
+	writeReport := func(format string, args ...any) {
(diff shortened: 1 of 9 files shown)


scfw/internal/cli/doctor.go
```diff
@@ -0,0 +1,94 @@
...
+		}
+		for name := range aliases {
+			aliasLocations[name] = append(aliasLocations[name], path)
+		}
+	}
+
+	for _, name := range availableAliases {
+		locations := aliasLocations[name]
+		if len(locations) == 0 {
+			fmt.Fprintf(cmd.OutOrStdout(), "❌ Alias %s is not set\n", name)
```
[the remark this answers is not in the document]
marcwieserdev: Intended, a user which run that command is likely to have something not working so it needs to be clear what is broken / unset

scfw/internal/cli/doctor.go
```diff
@@ -0,0 +1,94 @@
...
+	RunE:  runDoctor,
+}
+
+func runDoctor(cmd *cobra.Command, args []string) error {
+	// Installation and configuration checks will be added here.
+	apiKey, appKey, apiKeyCredentialSource, appKeyCredentialSource, credentialErr := ddapi.DDCredentials()
+
+	if credentialErr != nil {
+		fmt.Fprintf(cmd.OutOrStdout(), "❌ Credentials not found with an error : %v\n", credentialErr)
+	} else {
```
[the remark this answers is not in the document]
marcwieserdev: would be interesting to be able to distinguish errors to avoid repeated commands yes

scfw/internal/ddapi/auth.go
```diff
@@ -30,31 +30,57 @@ const (
...
 func ddCredentials() (ddAPIKey, ddAppKey string, err error) {
-	if ddAPIKey, err = resolveCredential(KeychainAPIKeyAccount, ddAPIKeyVar); err != nil {
-		return "", "", fmt.Errorf("failed to resolve Datadog API key: %w", err)
+	if ddAPIKey, _, err = ResolveDatadogAPIKey(); err != nil {
+		return "", "", err
 	}
-	if ddAppKey, err = resolveCredential(KeychainAppKeyAccount, ddAppKeyVar); err != nil {
-		return "", "", fmt.Errorf("failed to resolve Datadog application key: %w", err)
+	if ddAppKey, _, err = ResolveDatadogAppKey(); err != nil {
+		return "", "", err
```
dastrong, line 50: Should we use the `source` that's returned to detect mismatches (api key is from env; app key is from keychain)?

dastrong (approved):
Thanks for this 🙇 

Left a non-blocking comment

scfw/internal/ddapi/auth.go
marcwieserdev: Well I won't do that at first because it is technically a valid config (so it will be noted as valid) and the source will show up, so we have a way to check for this already, but I guess we will have our answer as soon as we try to onboard people :D

dastrong: I was thinking of a debug log here that notes a mismatch for quick debugging

PR #312, opened 2026-09-03
Remove findings field from Code Security API report payload

dastrong:
## 🚀 Motivation
`findings` is redundant now that the CLI reports `policies` and `failures` separately, and the field can be dropped.

## 📝 Summary
Removes the `findings` field from the Code Security API report payload emitted by the supply chain firewall CLI, since matched policies and verifier failures already convey that information.

## 🧪 Testing
- [x] Existing tests were updated for new logic, and not only so that they pass!
- [ ] New tests were added for new logic.
- [ ] Benchmark results prove that performance is the same or better.


piloulacdog (approved):
LGTM

